PR #8, opened 2016-10-08
Restore --with-x11 support

syl20bnr:
Not sure if the motivation of Homebrew maintainers is well funded but we don't care since we have our own tap, here is the x11 support! :-)

### DIFF
--- a/Formula/emacs-plus.rb
+++ b/Formula/emacs-plus.rb
@@ -29,11 +29,17 @@ class EmacsPlus < Formula
   deprecated_option "with-d-bus" => "with-dbus"
 
   depends_on "pkg-config" => :build
+  depends_on :x11 => :optional
   depends_on "dbus" => :optional
   depends_on "gnutls" => :recommended
   depends_on "librsvg" => :recommended
   depends_on "imagemagick" => :recommended
   depends_on "mailutils" => :optional
+
+  if build.with? "x11"
+    depends_on "freetype" => :recommended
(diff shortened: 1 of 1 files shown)


d12frosted:
Hm... Have you seen following error? 

```
==> ./configure --disable-silent-rules --enable-locallisppath=/usr/local/share/emacs/site-lisp --infodir=/usr/local/Cellar/emacs-plus/25.1/share/info/emacs --
Last 15 lines from /Users/d12frosted/Library/Logs/Homebrew/emacs-plus/01.configure:
checking for png... yes
checking whether png_longjmp is declared... yes
checking tiffio.h usability... yes
checking tiffio.h presence... yes
checking for tiffio.h... yes
checking for TIFFGetVersion in -ltiff... yes
checking gif_lib.h usability... no
checking gif_lib.h presence... no
checking for gif_lib.h... no
configure: error: The following required libraries were not found:
     libgif/libungif
Maybe some development libraries/packages are missing?
If you don't want to link with them give
     --with-gif=no
as options to configure

READ THIS: https://git.io/brew-troubleshooting
```

After I `brew install giflib` everything worked. Probably it should be added as dependency as well. Just to confirm, do you have `giflib` installed?

d12frosted:
Oh. When I start emacs, I see following error.

```
λ emacs

(process:38230): GLib-GIO-CRITICAL **: g_settings_schema_source_lookup: assertion 'source != NULL' failed
```

Issue #13, opened 2017-01-14
Borderless feature


kszucs:
I've applied the patch in the borderlass branch:
```bash
$ brew install --HEAD ./emacs-plus.rb                                                                                                                                              master
==> Cloning https://github.com/emacs-mirror/emacs.git
Updating /Users/krisz/Library/Caches/Homebrew/emacs-plus--git
==> Checking out branch master
==> Downloading https://lists.gnu.org/archive/html/bug-gnu-emacs/2016-10/binh950gCIBLY.bin
Already downloaded: /Users/krisz/Library/Caches/Homebrew/emacs-plus--patch-184301d28dabab5e2115cfd3c4875b89b408c31845ef04f909a6e0710523f082.bin
==> Patching
==> Applying binh950gCIBLY.bin
patching file lisp/cus-start.el
Hunk #1 succeeded at 448 (offset 2 lines).
patching file src/nsterm.m
Hunk #1 succeeded at 6748 (offset -27 lines).
Hunk #2 succeeded at 6787 (offset -27 lines).
Hunk #3 succeeded at 6809 (offset -27 lines).
Hunk #4 succeeded at 8736 with fuzz 1 (offset -27 lines).
==> ./autogen.sh
==> ./configure --disable-silent-rules --enable-locallisppath=/usr/local/share/emacs/site-lisp --infodir=/usr/local/Cellar/emacs-plus/HEAD-877c525/share/info/emacs --prefix=/usr/local/Cellar/emacs-plus/HEAD-877c525 --with-xml2 --without-d
==> make
==> make install
...
```

 but there is no `ns-use-titled-windows` variable in Emacs.

Could You give me some hints to achieve borderless frame? 

I've found another patch which conflicts with emacs' source http://stackoverflow.com/questions/24021624/in-gnu-emacs-osx-how-to-hide-title-bar .

d12frosted:
I have created that branch to share the work on this feature, but I couldn't make it working, that's why it's still not in `master` branch. The patch I have applied is taken from [this thread][1]. 

Wow, thanks for the link. I'll try it. But in case you wish to bring this patch faster, I highly advice you to test it on your own and send me a PR. This would be very nice and helpful as I am not going to look into it for several weeks. 

  [1]: https://lists.gnu.org/archive/html/bug-gnu-emacs/2016-10/msg00072.html

braham-snyder:
[...]

PR #14, opened 2017-01-16
Add compatible borderless (no title bar) patch and build option.

braham-snyder:
Note that this patch does not currently succeed on HEAD:
accordingly, throw an error if both --borderless-variable and --HEAD
are given.


d12frosted:
Thank you! I will test it soon and if it works, I'll merge it 😸

d12frosted:
Fixes #13

d12frosted:
I think it makes sense to leave it by default set to `nil`. 👍 

Thank you :+1: Squashed last two commits and cherry-picked into `master`, you can safely delete your branch.

PR #16, opened 2017-01-21
Added option to use multicolor font patch.

### LINKED ISSUE #15: Re-enabling multicolor font support in Emacs 25.1
As you may or may not be aware, in Emacs 25.1, the FSF made the decision to remove multicolor font support from the macOS version of Emacs because it is not currently supported by GNU/Linux. I just recently discovered this through [this blogpost](http://www.lunaryorn.com/posts/bye-bye-emojis-emacs-hates-macos.html) and now understand why my wasted hours of trying to get multicolor fonts working were just banging my head into a wall.

I'm not sure if you have any interest in changing this in your formula, but in order to resolve the issue personally, I prepared a patch that I used locally with this formula to get working multicolor font support once again. So, I thought I would bring this up here with [the patch](https://gist.github.com/aatxe/260261daf70865fbf1749095de9172c5), in case you wish to integrate it. I would make a PR but my local copy is using homebrew's embedded patch feature. That being said I can also prepare one if you're interested in adding this.

aatxe:
This resolves #15 by adding a `--with-multicolor-fonts` option which applies a patch reverting the commit that disabled multicolor font support for macOS.

### DIFF
--- a/Formula/emacs-plus.rb
+++ b/Formula/emacs-plus.rb
@@ -30,6 +30,7 @@ class EmacsPlus < Formula
   option "without-modules", "Build without dynamic modules support"
   option "without-spacemacs-icon", "Build without Spacemacs icon by Nasser Alshammari"
   option "with-ctags", "Don't remove the ctags executable that Emacs provides"
+  option "with-multicolor-fonts", "Build using patch for enabling multicolor font support."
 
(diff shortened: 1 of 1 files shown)


d12frosted:

I've enabled this patch by default (it seems useful for everyone) and updated docs a bit in a2d591b.

PR #22, opened 2017-03-24
Provide checksum for homebrew to be happy and change EOL encodings for

jwintz:
diff hunks to work.


d12frosted:

PR #23, opened 2017-03-24
Change option name based on the actual behavior of the API.

jwintz:


d12frosted:

jwintz:
Actually everything works fine. I had trouble since:

- I `osascript -e 'tell application "Finder" to make alias file to POSIX file "/usr/local/opt/emacs/Emacs.app" at POSIX file "/Applications"'`
- I have a transparent parameter that did not let me guess that the title bar color was inferred

I will leave the patch as it is since it works and since the behaviour is better than expected (hence the option name).

If anyone has a clue on why I get this middle line error message, please let me know.

PR #36, opened 2017-05-07
Update title bar patch sha256

aaronjensen:
The patch changed slightly. It doesn't actually work on 25.2 as far as I can tell, so it still needs some work to actually be functional now that this repo points to 25.2.

In other words, this may not be worth accepting. I may have some time tomorrow to get the patch working w/ 25.2


aaronjensen:
superceded by #37

PR #47, opened 2017-08-01
Fix README's link to issues page.

orther:
The existing link to the Issues page at the bottom of the README took you to issues page w/ no results because the query string was double url encoded. I just removed the query string all together because as far as I know github defaults to open issues.

I clicked this link a few times tonight and figured I'd save myself and anyone else a few seconds by making this pull request.

### DIFF
--- a/README.org
+++ b/README.org
@@ -75,7 +75,7 @@ with any arising issues.
 
 ** Known Issues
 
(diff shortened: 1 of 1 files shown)


d12frosted:

d12frosted:
Indeed that url was... bad 😸 thanks for noticing it and fixing that!

PR #58, opened 2017-12-05
Added a patch from emacs-mac for pixel scrolling.

aatxe:
One of the features that the emacs-mac distribution brags about is pixel scrolling when using the mouse wheel. It's behavior is slightly funky (from my expectation) in that it will still always stop line by line, but the actual scrolling process feels more like native Mac scrolling. I've added it as an option here. It's a pretty opinionated patch though, so I did not include it by default.

### DIFF
--- a/Formula/emacs-plus.rb
+++ b/Formula/emacs-plus.rb
@@ -32,6 +32,8 @@ class EmacsPlus < Formula
          "Don't remove the ctags executable that Emacs provides"
   option "without-multicolor-fonts",
          "Build without a patch that enables multicolor font support"
+  option "with-pixel-scrolling",
+         "Build with a patch from emacs-mac supporting native pixel scrolling"
   option "with-x11",
          "Experimental: build with x11 support"
   option "with-no-title-bars",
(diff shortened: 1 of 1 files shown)


d12frosted:
I am very open to new options 😸 Even unstable ones. Even if they are used only by one person. Have fun and thank you :+1: Cherry-picked into `master`, you can safely delete your branch.

aatxe:
Thanks! 🍻

aatxe:
I also tried out emacs-mac yesterday, and the behavior of emacs-plus with this patch is actually different from emacs-mac. They seem to have made changes to the renderer as well because they're able to have the scrolling "settle" mid-line. So, the pixel scrolling is even nicer there (and is more-or-less exactly what I wanted). On the other hand, they don't support native fullscreen or terminal UI. I'll try to see if we can get true pixel scrolling a la emacs-mac, but I'm leaving this info here for others too.

d12frosted:
Yeah, it seems that `emacs-mac` has many changes on different levels.

Thanks for heads up, though!

PR #62, opened 2018-01-18
Known issue note about rust dependency and compile time

johnbender:
Hi!

I was following the instructions for spacemacs to install `emacs-plus` and I noticed that the install was taking a while because it was installing rust. I inspected the dep tree with `brew deps --tree --include-build emacs-plus` and I saw that the from-source rust compiler package was a dependency of `librsvg` which can be disabled optionally.

Anyway, I thought I would add a note for others looking around and wondering why the rust compiler is required for this package. 

You might also consider making `librsvg` a `--with` dependency?

Oh, and don't feel obligated to merge this, I just tend to submit PRs instead of issues where there is some trivial solution available.

Thanks!

### DIFF
--- a/README.org
+++ b/README.org
@@ -83,6 +83,12 @@ with any arising issues.
 
 Please checkout [[https://github.com/d12frosted/homebrew-emacs-plus/issues][Issues]] page for a list of all known issues.
 
+*** Rust Dependency 
+	
+The librsvg dependency requires the rust compiler at build time. Rust can take
+quite a while to build (30+ minutes), so if you don't need SVG rendering support
+use the `--without-librsvg` flag when installing.
+
 ** Screenshots
 
 #+BEGIN_HTML


d12frosted:
Hey,

Thanks for reaching me out 😸 

I am really amazed that it takes so long to build (30+ minutes?).

The main problem is that I haven't updated to High Sierra yet, so I am not able to provide bottles for it. And I am delaying the update for a very long time. It would be super nice if someone (you?) could help me with bottles for High Sierra.

> You might also consider making librsvg a --with dependency?

Sorry, but no. I always advise people to build *with* `librsvg` and the idea of `emacs-plus` to provide most of the useful stuff out of the box without the need to specify any additional options.

johnbender:
@d12frosted 

Hey!

> I am really amazed that it takes so long to build (30+ minutes?).

Rust takes a very long time to compile.

> Sorry, but no.

I totally understand. Either way, hopefully some of this will be indexed for people who don't want rust and don't want to wait.

Thanks for responding!

aatxe:
I'm confused by this. @johnbender are you building the Rust compiler in order to build librsvg? There are binary installations of rustc/cargo.

d12frosted:
Ok, I am closing this issue as everyone is free to choose how to get dependencies. This PR can be found using search/google, so that's enough I think.

Thanks everyone.

PR #74, opened 2018-04-10
Updated --devel to the latest release candidate

dmgm:
The latest release candidate is supposed to be very ready for testing, with a 26.1 release the next step.
https://git.savannah.gnu.org/cgit/emacs.git/tree/etc/NEWS.26

I've built this locally and it works as well as the previous --devel build.

### DIFF
--- a/Formula/emacs-plus.rb
+++ b/Formula/emacs-plus.rb
@@ -10,8 +10,8 @@ class EmacsPlus < Formula
   end
 
   devel do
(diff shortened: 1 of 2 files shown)


dmgm:
So how about it?

d12frosted:
Holy crap! I didn't notice your PR. Sorry about that! 🙏

d12frosted:
Ok, using the web interface for such things is not a good option. Thanks for your commit, I've merged it into master ♥️

PR #78, opened 2018-05-24
apply wait_reading_process_ouput patch

d12frosted:
just wip, but I want to run it on CI


d12frosted:
It seems that the patch can't be applied clearly on master. I'll investigate a little bit later. But probably I won't spend much effort on it as 26.1 will be out soon.

PR #91, opened 2018-06-18
Fix spelling in warning messages

ksexton:


d12frosted:
Funny. Thanks for the fix! ❤️

PR #93, opened 2018-06-26
Update no title bars flag

floscr:
The `--with-no-titlebar` flag seems to be replaced with `--with-no-title-bars`

### DIFF
--- a/README.org
+++ b/README.org
@@ -88,7 +88,7 @@ relevant part from output of that command for your convenience.
     Build with mailutils support
   --with-modern-icon
     Using a modern style Emacs icon by @tpanum
-  --with-no-titlebar
+  --with-no-title-bars
     Experimental: build without titlebar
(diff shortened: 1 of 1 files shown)


d12frosted:
Hey @floscr,

Thank you for your PR.

Are you using `emacs-plus@25`? If so, then you are right. But I would rather say that `--with-no-title-bars` was renamed to `--with-no-titlebar` in `emacs-plus`:

```
  option "with-no-titlebar", "Experimental: build without titlebar"
  deprecated_option "with-no-title-bars" => "with-no-titlebar"
```

floscr:
Ah ok thanks! I was just guessing that users always would install the latest version 🙇

d12frosted:
> I was just guessing that users always would install the latest version

This is how it works - the latest. And in the latest version, it's `with-no-titlebar` 😸

PR #102, opened 2018-09-29
Remove ns-drawing patches

aaronjensen:
These are now in `emacs-26` and `master`

### DIFF
--- a/Formula/emacs-plus.rb
+++ b/Formula/emacs-plus.rb
@@ -80,16 +80,6 @@ class EmacsPlus < Formula
     if build.with? "xwidgets"
       odie "--with-xwidgets is not supported on Mojave yet"
     end
-
-    patch do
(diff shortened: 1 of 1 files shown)


d12frosted:
I didn't find any information (maybe I just missed it) in the Emacs devel list, but will there be a new Emacs 26 release with this fix?

shuxiao9058:
see https://github.com/emacs-mirror/emacs/commit/6217746dd64b43a2a2b3b66ab50cfbbfc984f36c

d12frosted:
@shuxiao9058 by release I meant something like https://ftp.gnu.org/gnu/emacs/

shuxiao9058:
Yes, the tarball is not update yet.

d12frosted:
I've checked it locally. Works as expected 😸 Thank you @aaronjensen !

I will deal with other options tomorrow.

PR #104, opened 2018-10-04
 apply Mojave fixes to emacs-26

d12frosted:


d12frosted:
CI passed and works locally.

Issue #117, opened 2019-02-28
Switch to ImageMagick 7 for HEAD


slippycheeze:
HEAD now supports ImageMagick 7; it'd be nice to enable that in the formula.

daviderestivo:
I have added ImageMagick 7 support to my personal homebrew GNU/Emacs formula. 
@d12frosted if you like you can use the code in [this commit](
https://github.com/daviderestivo/homebrew-emacs-head/commit/c437dce1b7b35cd56c2efff3022e23b850dcafdb).

d12frosted:
Merged to master.

P. S. @daviderestivo Sorry, I missed your message :( But the work got to this repo!

daviderestivo:
Thanks :)

PR #120, opened 2019-03-28
add option --with-imagemagick@7 and --with-jansson for HEAD

shuxiao9058:
Please have a test.

### DIFF
--- a/Formula/emacs-plus.rb
+++ b/Formula/emacs-plus.rb
@@ -66,6 +66,8 @@ class EmacsPlus < Formula
          increasedremembered_data size (--HEAD only)"
   option "with-xwidgets",
          "Experimental: build with xwidgets support (--HEAD only)"
+  option "with-jansson",
+         "Build with jansson support (--HEAD only)"
 
   # Disable some experimental stuff on Mojave
   if MacOS.full_version >= "10.14"
@@ -107,14 +109,18 @@ class EmacsPlus < Formula
   depends_on "dbus" => :optional
   depends_on "gnutls" => :recommended
   depends_on "librsvg" => :recommended
-  # Emacs 26.x does not support ImageMagick 7:
-  # Reported on 2017-03-04: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=25967
-  depends_on "imagemagick@6" => :recommended
-  # Emacs 27.x (current HEAD) does support ImageMagick 7:
-  # Prefer ImageMagick 7 when both on
-  depends_on "imagemagick@7" => :optional
 
   depends_on "mailutils" => :optional
+
+  if build.head?
+    # Emacs 27.x (current HEAD) does support ImageMagick 7
+    depends_on "imagemagick@7" => :recommended
+    depends_on "imagemagick@6" => :optional
+  else
+    # Emacs 26.x does not support ImageMagick 7:
+    # Reported on 2017-03-04: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=25967
+    depends_on "imagemagick@6" => :recommended
+  end
 
   depends_on "jansson" => :optional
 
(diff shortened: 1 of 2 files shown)


d12frosted:
Thank you very much! Only one test failed (and due timeout). I've restarted the job. Let's see how it goes.

Formula/emacs-plus.rb
```diff
@@ -233,12 +238,34 @@ def install
...
       args << "--without-imagemagick"
     end
 
+    # Emacs 27.x (current HEAD) supports imagemagick7 but not Emacs 26.x
+    if build.with? "imagemagick@7"
+      imagemagick_lib_path =  Formula["imagemagick@7"].opt_lib/"pkgconfig"
+      unless build.head?
+        odie "--with-imagemagick@7 is supported only on --HEAD"
+      end
+        ohai "ImageMagick PKG_CONFIG_PATH: ", imagemagick_lib_path
```
d12frosted: I think something happened to indentation here.

Formula/emacs-plus.rb
```diff
@@ -233,12 +238,34 @@ def install
...
+        ENV.prepend_path "PKG_CONFIG_PATH", imagemagick_lib_path
```
d12frosted: Wow, this thing looks so hacky. Never used it before :)

Formula/emacs-plus.rb
```diff
@@ -107,11 +107,17 @@ class EmacsPlus < Formula
...
+  # Emacs 26.x does not support ImageMagick 7:
   # Reported on 2017-03-04: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=25967
   depends_on "imagemagick@6" => :recommended
+  # Emacs 27.x (current HEAD) does support ImageMagick 7:
+  # Prefer ImageMagick 7 when both on
+  depends_on "imagemagick@7" => :optional
+
   depends_on "mailutils" => :optional
 
+  depends_on "jansson" => :optional
```
d12frosted: Once 27 is out this one should become recommended!

d12frosted:
Yay. Tests are green. I would like to play around with it. Hopefully I will have some time on weekends.

Meanwhile, it would be really nice if you could check if it's possible to manipulate the list of recommended dependencies based on the `--HEAD` option. If it's impossible, I am OK with it.

shuxiao9058:
the code refer to https://github.com/d12frosted/homebrew-emacs-plus/issues/117

shuxiao9058:
> I wonder if it's possible to actually make `imagemagick@7` recommended on `HEAD` and optional on stable using ifs.

I will try later today.

shuxiao9058:
@d12frosted I have a problem, how to show the option with imagemagick@7 and imagemagick@6?


``` ruby
  if build.head?
    # Emacs 27.x (current HEAD) does support ImageMagick 7:
    # Prefer ImageMagick 7 when both on
    depends_on "imagemagick@7" => :recommended
    depends_on "imagemagick@6" => :optional
  else
    # Emacs 26.x does not support ImageMagick 7:
    # Reported on 2017-03-04: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=25967
    depends_on "imagemagick@6" => :recommended
  end
```

With the code, the option only
```
--without-imagemagick@6
	Build without imagemagick@6 support
```

d12frosted:
Well, it makes sense. Let's not do it. Please update the README (section with options) and I am ready to merge :)

Formula/emacs-plus.rb
shuxiao9058: > Once 27 is out this one should become recommended!

It seems that the native json support is on with default.

 https://github.com/emacs-mirror/emacs/blob/eb6bbd9fb175cacdfdc54c1187f5785ed3858f2f/configure.ac#L432

d12frosted:
Thank you for your PR :heart: :100: It's not in master, so you can safely remove your branch.

P. S. I've actually squashed your commits and fixed conflict. Resulting commit is 368431aceec319b56c90b63ea830216d16c503a0.

Issue #119, opened 2019-03-16
Disable focusing other frame when closing one


xenodium:
When closing a frame, Emacs automatically focuses another frame.

![before](https://user-images.githubusercontent.com/8107219/54477896-51be3600-4804-11e9-800a-9b25b9611ce7.gif)

This re-focus has an additional side-effect: when closing a frame from one desktop/space, one gets automatically moved to another desktop/space where the refocused frame lives.

Not been able to find a way to disable the re-focusing in elisp. However, a [small patch in frame.c](https://github.com/xenodium/homebrew-emacs-plus/commit/79a8e98b222fa262746f6db620ddd1fdb151ece9) seems to have done the job:

![after](https://user-images.githubusercontent.com/8107219/54477966-1708cd80-4805-11e9-9bfc-c64b71001be6.gif)

Patched in a fork. Any interest in including this option (disable refocus) in emacs-plus upstream recipe? Happy to submit a pull request.

d12frosted:
I see 3 thumb-ups. I say - send PR, I will be glad to merge it. :smile_cat:

d12frosted:
Merged to master.

PR #122, opened 2019-03-29
Disable focusing on other frame when closing current one

xenodium:
When closing a frame, Emacs automatically focuses another frame. This re-focus has an additional side-effect: when closing a frame from one desktop/space, the user gets automatically moved to another desktop/space where the refocused frame lives. This can be distracting.

Pull request disables this behavior by introducing the --with-no-frame-refocus install option and thus fixing #119.


d12frosted:
Thank you for your PR :heart: :100: It's not in master, so you can safely remove your branch.

P. S. Kudos for signing!

PR #126, opened 2019-04-13
Feature/26.2

shuxiao9058:


shuxiao9058:
install  emacs 26.2 with `brew install https://raw.githubusercontent.com/shuxiao9058/homebrew-emacs-plus/feature/26.2/Formula/emacs-plus.rb`

Formula/emacs-plus.rb
```diff
@@ -85,8 +85,8 @@ class EmacsPlus < Formula
   end
 
   devel do
-    url "https://alpha.gnu.org/gnu/emacs/pretest/emacs-26.1-rc1.tar.xz"
-    sha256 "6594e668de00b96e73ad4f168c897fe4bca7c55a4caf19ee20eac54b62a05758"
```
d12frosted: This is something I forgot to remove :smile_cat: Since 26.2 is released, I think we should remove `devel` option until new RC is available.  

P. S. I will remove it, so don't worry.

d12frosted:

PR #127, opened 2019-04-17
remove devel option

d12frosted:
cc @shuxiao9058


shuxiao9058:
you should change patch url  https://github.com/d12frosted/homebrew-emacs-plus/blob/8257cd367d97b3bfddfdfb0eb75c3b49ae402382/Formula/emacs-plus.rb#L135 to this repo.

d12frosted:
Oh right. I missed that. Will be available in two hours.

PR #129, opened 2019-04-17
update borderless-frame-on-macOS.patch url

shuxiao9058:


d12frosted:
You are fast! Thanks, will merge once I am home.

d12frosted:

Issue #130, opened 2019-04-19
UX: frame doesn't fill window space


winmillwill:
I found this project while looking for a build with no titlebar, and that seems to work fine, and so does everything else, so kudos and thank you!

One random gotcha that may be worth documenting is that if you don't have something like `(setq frame-resize-pixelwise t)` then you're liable to have some wonky gaps between windows when tiling, unless your typeface is somehow exactly aligned with your effective display area.

d12frosted:
Hi @winmillwill ,

Thank you for the feedback. I am glad that it works for you. 

Indeed, makes sense to document this. PR would be welcome :smile_cat:

daviderestivo:
> One random gotcha that may be worth documenting is that if you don't have something like `(setq frame-resize-pixelwise t)` then you're liable to have some wonky gaps between windows when tiling, unless your typeface is somehow exactly aligned with your effective display area.

Thanks a lot for sharing. I was not aware about this option.

d12frosted:
Documented thanks to @winmillwill  :smile_cat:

PR #132, opened 2019-04-22
Document frame resize behavior for tiling

winmillwill:
This is meant to address #130.

Open to suggestions as to a better place to put this.


d12frosted:

PR #134, opened 2019-04-29
Only apply multicolor-font patch if not on HEAD

majorgreys:
Given its inclusion in [a recent change][1], we no longer need to patch the `src/macfont.m` file if installing from `HEAD`.

[1]: https://github.com/emacs-mirror/emacs/commit/28220664714c50996d8318788289e1c69d69b8ab#diff-ee87d80780a3bc650c7aedfb3c1bd372


d12frosted:
These are some cool news! Thank you!

PR #136, opened 2019-05-14
Update xwidgets patch to a newer one

pcr910303:
Fixes #94 👍

PR #138, opened 2019-05-16
Add updated xwidget patch and modify formula accordingly

pcr910303:
Sorry for #136 :-(

I reset the repo to HEAD~3 and re-added the (right) patch.
As before, the patch is based on your repo, which means the formula won't work until this PR becomes merged.

I modified Homebrew's cache directory to the right file and it looks like it's working.


d12frosted:
Rebased on top of master and merged. I will test it today :)

d12frosted:
P. S. hash is right now :)

PR #139, opened 2019-05-16
Add updated xwidget patch and modify formula accordingly

pcr910303:
Sorry for #138 :-(

I reset the repo to HEAD~3 and re-added the (right) patch.
As before, the patch is based on your repo, which means the formula won't work until this PR becomes merged.


d12frosted:
Rebased and merged :)

Issue #146, opened 2019-07-14
Patch failed with recent commits.


smallzhan:
`brew install emacs-plus --HEAD --with-xwidgets --with-jansson --without-spacemacs-icon`
with the commit id `bcc3d36d8a`.

The patch `xwidgets_webkit_in_cocoa.patch` failed on file `src/xwidget.c` 

>patching file src/xwidget.c
Hunk \# 1 FAILED at 18.
Hunk \# 2 succeeded at 67 (offset 1 line).
Hunk \# 3 succeeded at 81 (offset 1 line).
Hunk \# 4 succeeded at 99 (offset 1 line).
Hunk \# 5 succeeded at 117 (offset 1 line).
Hunk \# 6 succeeded at 176 (offset 1 line).
Hunk \# 7 succeeded at 214 (offset 1 line).
Hunk \# 8 succeeded at 253 (offset 1 line).
Hunk \# 9 succeeded at 269 (offset 1 line).
Hunk \# 10 succeeded at 305 (offset 1 line).
Hunk \# 11 succeeded at 563 (offset 1 line).
Hunk \# 12 succeeded at 573 (offset 1 line).
Hunk \# 13 succeeded at 587 (offset 1 line).
Hunk \# 14 succeeded at 645 (offset 1 line).
Hunk \# 15 succeeded at 661 (offset 1 line).
Hunk \# 16 succeeded at 736 (offset 1 line).
Hunk \# 17 succeeded at 754 (offset 1 line).
Hunk \# 18 FAILED at 777.
Hunk \# 19 succeeded at 846 (offset 2 lines).
Hunk \# 20 succeeded at 885 (offset 2 lines).
Hunk \# 21 succeeded at 929 (offset 2 lines).
Hunk \# 22 succeeded at 940 (offset 2 lines).
Hunk \# 23 succeeded at 951 (offset 2 lines).
Hunk \# 24 succeeded at 972 (offset 2 lines).
Hunk \# 25 succeeded at 981 (offset 2 lines).
Hunk \# 26 succeeded at 991 (offset 2 lines).
Hunk \# 27 succeeded at 1017 (offset 2 lines).
Hunk \# 28 succeeded at 1100 (offset 2 lines).
Hunk \# 29 succeeded at 1218 (offset 2 lines).
Hunk \# 30 succeeded at 1392 (offset 2 lines).
Hunk \# 31 succeeded at 1421 (offset 2 lines).
[...]

PR #149, opened 2019-07-27
Update xwidgets_webkit_in_cocoa.patch

fuxialexander:
Resolve conflict on HEAD

### DIFF
--- a/Formula/emacs-plus.rb
+++ b/Formula/emacs-plus.rb
@@ -147,7 +147,7 @@ class EmacsPlus < Formula
     else
       patch do
         url "https://raw.githubusercontent.com/d12frosted/homebrew-emacs-plus/master/patches/xwidgets_webkit_in_cocoa.patch"
-        sha256 "bd614add2ca3b0da940521116b8b78f605213229747502c2159a728db5ac82d3"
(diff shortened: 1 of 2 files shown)


fuxialexander:
Should solve https://github.com/d12frosted/homebrew-emacs-plus/issues/146
Not tested on pdumper branch, but since it's already merged to master I guess it's fine.

zachdaniel:
Could this be merged? If so, that would definitely save me some time :)

d12frosted:
Sorry for the delay with this patch.

Two comments:

1. SHA is not the right one, should be: `bd614add2ca3b0da940521116b8b78f605213229747502c2159a728db5ac82d3`.
2. It doesn't apply on master cleanly.

@fuxialexander how is it going with the patch on the upstream? :)

JSpenced:
Is there a fix for this coming? Also, when I open an org-file it just freezes and I am guessing might be due to a rendering problem. Is there a way to install previous heads? I forgot to pin the package in brew so it updated.

xeijin:
@fuxialexander any luck with this one?

d12frosted:
Build was fixed in #166. So closing this.

PR #160, opened 2019-12-19
[WIP] Add no-titlebar support for --HEAD

jakejx:
Hi,

I have updated the no-titlebar patch to work with HEAD. Full credits go to the original author, I simply updated the diff for the correct lines. I have also attempted to fix the window role assigned to Emacs so that it is more well behaved with tiling managers such as yabai. I have only tested for abit but I am hoping that more people can test this build to make sure its working.

Let me know if there is any work to be done before merging, thanks! 😄

### DIFF
--- a/Formula/emacs-plus.rb
+++ b/Formula/emacs-plus.rb
@@ -113,12 +113,15 @@ class EmacsPlus < Formula
 
   if build.with? "no-titlebar"
     if build.head?
-      odie "--with-no-titlebar is not supported on --HEAD yet"
-    end
-
-    patch do
-      url "https://raw.githubusercontent.com/d12frosted/homebrew-emacs-plus/master/patches/borderless-frame-on-macOS.patch"
-      sha256 "2059213cc740a49b131a363d6093913fa29f8f67227fc86a82ffe633bbf1a5f5"
+      patch do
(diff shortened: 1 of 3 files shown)


d12frosted:
Nice! Thank you very much for putting your effort into it. I will check it on weekends. 

Would be nice if the patch itself was added to the patches directory in this repository. In the PR you can use a link to the file on master. Will work after merging :)

jakejx:
> Nice! Thank you very much for putting your effort into it. I will check it on weekends.
> 
> Would be nice if the patch itself was added to the patches directory in this repository. In the PR you can use a link to the file on master. Will work after merging :)

Ok sure thing I will add it in tonight. For testing reasons I had to use an external gist. As a side note, I might also split the patches into two distinct patches. One for --without-titlebar and one for fixing the window role.

d12frosted:
Sure, take your time. And I totally understand the testing needs :) You can leave the URL as is, I will fix that, but please attach the patch to PR.

Regarding the separation of patches - it makes a lot of sense.

jakejx:
I have included the patches and separated the patches. I have indicated that the window role patch is experimental. I have also updated the patch urls to use this repo.

I have tested locally and Emacs is building fine. However, I have not tested different combinations of options (there are so many) so I can't be certain that the patches will work with other options. Does the CI automatically try different combinations?

d12frosted:
Thank you very much!

PR #171, opened 2020-01-04
merge window-role-fix option

### LINKED ISSUE #157: Wrong window role
First of all thanks for this awesome distribution! One of the problems I faced was with the window role being `AXTextField`. This causes problems with the window tiler I use, as you can see at its issues: ([yabai](https://github.com/koekeishiya/yabai/issues?utf8=%E2%9C%93&q=is%3Aissue++emacs+)). 

I'm not knowledgeable, but it seems like the easiest change would be to change the Emacs window role to `AXWindow`.

d12frosted:
Fixes #157


d12frosted:
Merged to master

PR #200, opened 2020-04-18
update patches

d12frosted:


d12frosted:
Tested locally the no-title-bar patch and it works.

PR #206, opened 2020-04-23
Add system-appearance-change patch & build option

ngquerol:
Hello 👋🏻

I made this patch a while ago, and decided to make a PR here to gauge interest in it.
Everything is explained in the patch's description, but to summarize:

This adds a new hook, `ns-system-appearance-change-functions`, that is run every time the system appearance changes. Functions added to this hook will be called with one argument, a symbol that is either `'light` or `'dark`. This mainly allows loading a different theme to better match the system appearance.

See below for an example of its usage:

``` emacs lisp
(add-hook 'ns-system-appearance-change-functions
        #'(lambda (appearance)
            (mapc #'disable-theme custom-enabled-themes)
            (pcase appearance
               ('light (load-theme 'tango t))
               ('dark (load-theme 'tango-dark t)))))
```

I know that you are trying not to complicate this formula any further, so let me know if adding yet another build option for such a trivial feature is not welcome. Additionally the patch only applies to the `emacs-27` branch, but I could always try to make another one for `master` if need be.

### DIFF
--- a/Formula/emacs-plus.rb
+++ b/Formula/emacs-plus.rb
@@ -64,6 +64,8 @@ class EmacsPlus < Formula
          "Build from emacs-27-branch (--HEAD only)"
   option "with-native-comp-branch",
          "Build from native-comp branch (--HEAD only)"
+  option "with-system-appearance-change",
+         "Built with support for system appearance changes (--HEAD only)"
 
   # Update list from
   # https://raw.githubusercontent.com/emacsfodder/emacs-icons-project/master/icons.json
@@ -175,6 +177,12 @@ class EmacsPlus < Formula
     end
   end
 
+  if build.with? "system-appearance-change"
+    unless build.head?
+      odie "--with-system-appearance-change is supported only on --HEAD"
+    end
+  end
+
   #
   # Patches
   #
@@ -231,6 +239,20 @@ class EmacsPlus < Formula
   patch do
     url (PatchUrlResolver.url "fix-window-role")
     sha256 "f6cd00bcd37be03d8d83c2e590ab7d259ae37632f5954267240ba9439f799359"
+  end
+
+  if build.with? "system-appearance-change"
+    if build.with? "emacs-27-branch"
+      patch do
+        url (PatchUrlResolver.url "system-appearance-change-emacs-27")
+        sha256 "82252e2858a0eba95148661264e390eaf37349fec9c30881d3c1299bfaee8b21"
+      end
+    else
+      patch do
+        url (PatchUrlResolver.url "system-appearance-change")
+        sha256 "c9d3b0ed492c0e2aed6661c76efef3949e25b0c0297ddfb806b5d33639386eb5"
+      end
+    end
   end
 
   #
--- a/patches/system-appearance-change-emacs-27.patch
+++ b/patches/system-appearance-change-emacs-27.patch
@@ -1,0 +1,234 @@
+From 48f853d2deb2ddb283a64376f28d16bc033e634a Mon Sep 17 00:00:00 2001
+From: "Nicolas G. Querol" <nicolas.gquerol@gmail.com>
+Date: Wed, 18 Mar 2020 10:49:28 +0100
+Subject: [PATCH] Add `ns-system-appearance-change-functions' hook
+
+This implements a new hook, effective only on macOS >= 10.14 (Mojave),
+that is called when the system changes its appearance (e.g. from light
+to dark). Users can then implement functions that takes this change
+into account, for instance to load a particular theme.
+
+The frame parameter `ns-appearance', if set, will result in this new
+hook not being called until this parameter is unset.
+
+Minor changes are also made to select the right "dark" appearance
+(NSAppearanceNameDarkAqua) on macOS versions >= 10.14, the previous one
+(NSAppearanceNameVibrantDark) being deprecated.
+
+* src/frame.h (enum ns_appearance_type): Add new
+"ns_appearance_dark_aqua" case.
+
+* src/nsfns.m (defun x-create-frame): Use "dark aqua" appearance on
+macOS >= 10.14.
+
+* src/nsterm.m:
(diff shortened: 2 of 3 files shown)


Formula/emacs-plus.rb
```diff
@@ -233,6 +235,17 @@ class EmacsPlus < Formula
     sha256 "f6cd00bcd37be03d8d83c2e590ab7d259ae37632f5954267240ba9439f799359"
   end
 
+  if build.with? "system-appearance-change"
+    unless build.with? "emacs-27-branch"
```
d12frosted, line 239: There is a separate 'section'[1] for incompatible options. I find it easier to keep them in one place.

  [1]: https://github.com/d12frosted/homebrew-emacs-plus/pull/206/files#diff-cbfee31de48c4c925ef3e240a0362bb2R156

Formula/emacs-plus.rb
```diff
@@ -233,6 +235,17 @@ class EmacsPlus < Formula
...
+      odie "--with-system-appearance-change is supported only when building "\
+           "with --with-emacs-27-branch"
```
d12frosted, line 241: I understand why it should be `with <option>`, though if feels funny since most options already contain word `with`. Not sure what is the best way to word it haha.

patches/system-appearance-change-emacs-27.patch
```diff
@@ -0,0 +1,234 @@
...
++#define NSAppKitVersionNumber10_14 1671
++#endif
++- (void)viewDidChangeEffectiveAppearance
++{
++  NSTRACE ("[EmacsView viewDidChangeEffectiveAppearance:]");
++
++  if (NSAppKitVersionNumber < NSAppKitVersionNumber10_14)
++    return;
++
++  // The `ns-appearance' frame parameter overrides the system appearance;
```
d12frosted, line 181: Do I understand it correctly, that by default `ns-appearance` is not set, which means that with this patch, Emacs appearance will change between dark and light depending on the system configuration?

Formula/emacs-plus.rb
```diff
@@ -233,6 +235,17 @@ class EmacsPlus < Formula
...
+    end
+    patch do
+      url (PatchUrlResolver.url "system-appearance-change-emacs-27")
```
d12frosted, line 244: I fixed `PatchUrlResolver` yesterday for PRs and you are the first person to actually add a new patch via PR using this resolver. 💯 Happy to see that it works.

Formula/emacs-plus.rb
ngquerol: 👍🏻 I moved the check to that section.

Formula/emacs-plus.rb
ngquerol: I changed the phrasing, since I added a `master`version of the patch.

patches/system-appearance-change-emacs-27.patch
ngquerol: Yes; As long as the frame parameter `ns-appearance` is not explicitly set to `'dark` or `'light`, the appearance of the frame will follow the global system appearance, and the functions added to the new hook variable will be run.

Formula/emacs-plus.rb
ngquerol: It works perfectly indeed! 🥳

ngquerol:
Hello again ! I produced an additional version of the patch that applies cleanly to the `master` branch. Apologies for the delay, but there had been commits on that branch that impacted the same code as this patch, and I needed to think a little bit about how to adapt it.

All your remarks should have been addressed, that said I left the patch hidden behind the option `--with-system-appearance-change` for now.

d12frosted:
I've tested it with in the following configurations:

- `--HEAD --with-system-appearance-change`
- `--HEAD --with-system-appearance-change --with-emacs-27-branch`
- `--HEAD --with-system-appearance-change --with-no-titlebar`
- `--HEAD --with-system-appearance-change --with-emacs-27-branch --with-no-titlebar`

The feature works nicely! I love it (though I always use light mode of macos) and think many people will find it really useful. And it's not obtrusive at all, so I think it makes sense to enable it by default.

I haven't tested all invariants, but this patch doesn't apply when building `--with-no-titlebar`. Because of this, we should not make it unconditional. What do you think, is it possible to fix the patches so they work with `no-titlebar`? Ideally it should be one patch per branch (e.g. master vs emacs27), but if it's impossible, two will work as well.

d12frosted:
BTW, you can add more variants to the build matrix in the [workflow definition file][1].

  [1]: https://github.com/d12frosted/homebrew-emacs-plus/blob/master/.github/workflows/main.yml#L17

patches/system-appearance-change.patch
```diff
@@ -0,0 +1,219 @@
...
+     FRAME_NS_APPEARANCE (f) = ns_appearance_system_default;
++  }
+ 
+   [window setAppearance];
+ #endif /* MAC_OS_X_VERSION_MAX_ALLOWED >= 101000 */
+@@ -7595,7 +7604,6 @@ - (instancetype) initFrameFromEmacs: (struct frame *)f
+   if (! FRAME_UNDECORATED (f))
+     [self createToolbar: f];
+ 
+-
```
d12frosted, line 133: Ok, it seems that the conflict is trivial and all you need to do is to remove this line from patch. I will fix that by myself. Don't worry :)

d12frosted:
Once again, thank you very much for your PR. I've cherry-picked it to master (75c90fdd6cabf3965fb6fc3cdbe5dc0346b4a9ec). You can safely remove your branch.

I will also go an enable it by default.

d12frosted:
Your patch in [emacs-plus-basis][1] is added in [08a38fd7cf][2] 

  [1]: https://github.com/d12frosted/emacs-plus-basis
  [2]: https://github.com/d12frosted/emacs-plus-basis/commit/08a38fd7cfb250afd7b848aa68360202e3c180a7

ngquerol:
🙌 Thanks for the fix, and for maintaining this tap.

PR #207, opened 2020-04-27
apply system-appearance patch unconditionally

d12frosted:
P. S. the build may fail due to Xcode version.


d12frosted:
I will net merge until CI is fixed.

PR #208, opened 2020-04-30
Clarify "system-appearance" patch documentation

ngquerol:
👋🏻

Upon re-reading the documentation you added to the README concerning the `system-appearance` patch, I found some clarifications were needed:

- Emacs used to already react to system appearance changes if the `ns-appearance` frame parameter was not set; The titlebar does adopt the correct appearance in such situations, the patch however does provide the possibility to additionally run functions that take the appearance into account.
- Functions can be registered in `early-init.el`, so that Emacs directly adopts the correct theme upon startup
- The hook is not run in TTY Emacs sessions

I thought this would be useful to point out, in order to reduce potential confusion.

### DIFF
--- a/README.org
+++ b/README.org
@@ -160,12 +160,11 @@ More details can be seen here [[https://github.com/veshboo/emacs][Veshboo's emac
 
 *** System appearance change
 
-This patch is enabled by default and can't be disabled. First of all, it adds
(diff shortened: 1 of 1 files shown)


d12frosted:
Thank you very much, I've cherry-picked this commit (30d1b86).

PR #216, opened 2020-05-12
unconditionally provide support of certain features

d12frosted:
- gnutls
- librsvg
- libxml2
- little-cms2
- dynamic modules
- multicolor fonts

### DIFF
--- a/Formula/emacs-plus.rb
+++ b/Formula/emacs-plus.rb
@@ -38,14 +38,8 @@ class EmacsPlus < Formula
   # Opt-out
   option "without-cocoa",
          "Build a non-Cocoa version of Emacs"
-  option "without-libxml2",
-         "Build without libxml2 support"
-  option "without-modules",
-         "Build without dynamic modules support"
   option "without-spacemacs-icon",
          "Build without Spacemacs icon by Nasser Alshammari"
-  option "without-multicolor-fonts",
-         "Build without a patch that enables multicolor font support"
 
   # Opt-in
   option "with-ctags",
@@ -124,11 +118,13 @@ class EmacsPlus < Formula
   end
 
   depends_on "pkg-config" => :build
-  depends_on "little-cms2" => :recommended
+
+  depends_on "gnutls"
+  depends_on "librsvg"
+  depends_on "little-cms2"
+
   depends_on :x11 => :optional
   depends_on "dbus" => :optional
-  depends_on "gnutls" => :recommended
-  depends_on "librsvg" => :recommended
   depends_on "mailutils" => :optional
 
   if build.head?
@@ -190,12 +186,10 @@ class EmacsPlus < Formula
     end
   end
 
-  if build.with? "multicolor-fonts"
-    unless build.head?
-      patch do
-        url (PatchUrlResolver.url "multicolor-fonts")
-        sha256 "7597514585c036c01d848b1b2cc073947518522ba6710640b1c027ff47c99ca7"
-      end
+  unless build.head?
+    patch do
+      url (PatchUrlResolver.url "multicolor-fonts")
+      sha256 "7597514585c036c01d848b1b2cc073947518522ba6710640b1c027ff47c99ca7"
     end
   end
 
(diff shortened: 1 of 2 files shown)


d12frosted:
Let me share some rationale behind this change. 

I don't have options usage statistic (from the moral point of view, it's even good). But I don't think that someone really disables these things. I don't see how it can hurt someone having support for dynamic modules, rsvg or gnutls. If it does - please let me know and I will consider reverting some of the options. But keep in mind that my plan is to decrease amount of available options, so the formula is simpler to maintain.

cc @daviderestivo

cilia:
gnutls, librsvg, by default yes. But is it possible to have the option not to install libxml2, little-cms2, dynamic modules, and multicolor fonts?

d12frosted:
@cilia sure, can you please share reasons not to enable them?

cilia:
This might be opinionated but we're talking about Emacs anyway :)

- libxml2: how many people use it _on a Mac_ for things like eww or EPUB reader?
- multicolor fonts: isn't this removed for Mac by default because Unix/Linux doesn't support it yet? It may be included in future Emacs version but not in the current 26.3, and I guess not many people needs emoji in their Emacs.
- little-cms2: I assume this is related to multicolor fonts.
- dynamic modules: this is not enabled by default in official Emacs so why here, and certainly not a must for many (e.g. beginner/intermediate users, those who just want to use Emacs as a text editor).

My point is it'd be better to have options instead of being given a relatively bloated Emacs without choice. Thanks.

d12frosted:
> This might be opinionated but we're talking about Emacs anyway :)

Haha, indeed. It's Emacs land 😸 These kind of questions are always debatable.

/Apologies for the long answer/

I agree with all your points. Let me now share my perspective with one simple question - does it hurt anyone if these are enabled and can't be disabled? Apart from having a little bit longer compilation time (e.g. additional dependency and patch application), though the difference is so small that I don't even notice it.

Not having an option for these things makes it easier to maintain formula. Apart from `brew` disliking options and breaking them time to time, having lots of compilation options (especially patches) makes it much harder to ensure that all combinations work. I value fast CI and patch fixing ease. So my strategy right now is to enable unconditionally all 'extra' stuff that does not hurt regular users. Even if user doesn't plan to use it initially, he/she can change mind without the need to recompile Emacs (great example of such feature is [system appearance change hooks][1]).

On the contrary, if a feature changes Emacs behaviour drastically or have runtime tradeoffs, it does make sense leave decision to the user. Examples:

- `no-titlebar` changes UI, which can't be controlled via lisp AND it also breaks `yabai` users flow.
- `xwidgets` is a cool thing, but it's huge and it used to conflict time to time, so it's dangerous to unconditionally enable it. 
- `debug` affects runtime
- `x11` or `cocoa` or even nothing is a real choice that can be made only by users.

On the other hand, `mailutils`, `dbus` and `ctags` are optional. I am yet to better understand their implication on the user. 🤷 

I also think about breaking formula into 3: `emacs-plus`, which installs release version (e.g. 26), `emacs-plus@27` and `emacs-plus@28`. But this is topic for another thread haha. Unless someone wants me to elaborate.

> My point is it'd be better to have options instead of being given a relatively bloated Emacs without choice. Thanks.

I also do value this approach 😸 So here I am trying to balance common sense and maintenance cost. But I am not a tyrant, so if someone has strong feelings about any of the options, I can change my mind 😸 But I need elaboration.

I know that I've posted a wall of text, but I would really love to hear what you think @cilia ! 

  [1]: https://github.com/d12frosted/homebrew-emacs-plus/tree/b422ccfc820b542e1a8d505710566c1a03aa872c#system-appearance-change

cilia:
Thanks for the detailed feedback. I understand that as the main maintainer you want to make the quality and optionality of the package affordable, which mainly comes down to the cost/ease of maintaining. That's the balancing act you need to make, while we as users can provide feedback. And we thank you for your effort and time put into this.

Besides (somewhat subjective) justification for individual options/extensions, I think it's less debatable that we all come to love Emacs because it's an extensible and customizable piece of software such that people have the options to configure it in whatever way works best for them. I'll pause here and let others weigh in.

Issue #195, opened 2020-03-28
Broken installation of Emacs 26.3 when build from sources


albertparedandan:
**Describe the bug**

See [d12frosted's comment](https://github.com/d12frosted/homebrew-emacs-plus/issues/195#issuecomment-609978037) for more information.

When i install emacs using the without spacemacs icon argument the resulting emacs is not executable both from Applications folder and terminal. I get a zsh: killed emacs

I can't even run emacs--version or emacs doctor

**Command used to install `emacs-plus`**
```
$ brew install emacs-plus --without-spacemacs-icon
```

**Installation log**
Updating Homebrew...
==> Installing emacs-plus from d12frosted/emacs-plus
Warning: A newer Command Line Tools release is available.
Update them from Software Update in System Preferences or
https://developer.apple.com/download/more/.

==> Downloading https://ftp.gnu.org/gnu/emacs/emacs-26.3.tar.xz
Already downloaded: /Users/albertpare/Library/Caches/Homebrew/downloads/35ddbf65249f3d4c38e087a1fa7606daaf8784b8445fa2c3b4d453ef9b92a18c--emacs-26.3.tar.xz
==> Downloading https://raw.githubusercontent.com/d12frosted/homebrew-emacs-plus/master/patches/multicolor-fonts.patch
Already downloaded: /Users/albertpare/Library/Caches/Homebrew/downloads/6e0bcfe5f86b21ca9f1bf349995832a6143baac4a74e000dff771b2839b6fa7d--multicolor-fonts.patch
==> Downloading https://raw.githubusercontent.com/d12frosted/homebrew-emacs-plus/master/patches/fix-window-role.patch
Already downloaded: /Users/albertpare/Library/Caches/Homebrew/downloads/b7046a273c114e23f734df97967f4016486f1d080bad1ba40dbd53a5c86a09e0--fix-window-role.patch
==> Patching
==> Applying multicolor-fonts.patch
patching file src/macfont.m
Hunk #1 succeeded at 2346 with fuzz 2 (offset -27 lines).
==> Applying fix-window-role.patch
patching file src/nsterm.m
Hunk #1 succeeded at 8368 (offset -158 lines).
==> ImageMagick PKG_CONFIG_PATH:
/usr/local/opt/imagemagick@6/lib/pkgconfig
[...]

PR #226, opened 2020-06-04
split formula per Emacs version

d12frosted:
Just a proof of concept.

First thing to notice is that I've split the formula into 3 different formulas per Emacs version - `emacs-plus@26`, `emacs-plus@27` and `emacs-plus@28`. 

The plan is to have `emacs-plus` to build default/release version. And have separate formulas for any other versions.

I see few reasons to do it:

1. I see that some users are confused by all these versions (`--HEAD` vs `--HEAD --with-emacs-27-branch` vs no such options). And for some of them it's not clear that `emacs-plus` actually has three different versions.
2. Formula simplification. Just look at the `emacs-plus@26`, it's clear, though it doesn't support much (and due to #195 it doesn't support ANY options right now). In other versions there is no `if-elif-else` code anymore. Everything is straightforward. It is much easier to think about specific Emacs version. There are almost no `odie` calls.
3. Allows better control over options for specific version. For example, Emacs 27 actually has prebuilt `tar`, but it can also be built from `--HEAD` (e.g. `emacs-27` branch). 
4. Good side effect - better support of Emacs 28, no `--HEAD` anymore! Rebuilds via `brew reinstall` will work better, though I don't really care about them as they are broken on their very essence.

So to sum up, the split gives better user experience AND easier maintenance. On the other side, code and patch duplication. Which I can live with.

Other changes in this PR:

- Already mentioned, but almost all options are dropped from Emacs 26. Even icons. Unfortunately. I will prepare bottles so people can installed bottled version that works.
- Added prebuilt Emacs 27 in addition to `--HEAD`.

Tasks:

- [x] Update README.
- [x] Add a note that Emacs 26 is deprecated and that people should use Emacs 27 or Emacs 28. Can be done outside of this PR.
- [x] Rename `emacs-plus@26` to regular `emacs-plus`. I use other name just for testing purposes. Open for discussions. Maybe Emacs 27 should become default?
- [x] Fix `build` script.
- [x] Error out with message about new formulas when someone tries to install `emacs-plus` with `--HEAD` option.

---

Ok, it's kind of chaotic. Would be glad to hear what you folks think.

CC @daviderestivo


d12frosted:
Another valid question - how to 'migrate' users. Should we add a message like 'please use emacs-plus@28` if the user used `--HEAD` option?

d12frosted:
Wow. Something even builds 😅

d12frosted:
Found the issue with Emacs 28. Forgot to set `version` for the formula, as it can't be inferred anymore. 

```
Error: emacs-plus@28: version missing for "spacemacs-icon" resource!
```

The [error][1] I've got was super counterintuitive. 

Let's see how it goes.

P. S. I've also adapted the build script in 25d3e81. Now it accepts version argument. E.g. `./build 28 --with-no-titlebar`. 

  [1]: https://github.com/d12frosted/homebrew-emacs-plus/pull/226/checks?check_run_id=741080585#step:3:8

d12frosted:
[Ah right][1]. Now I remember why I didn't put `--with-x11` to the build matrix in the first place.

  [1]: https://github.com/d12frosted/homebrew-emacs-plus/pull/226/checks?check_run_id=741257250#step:3:9

d12frosted:
Hooray! I fixed the build. Next steps are to update documentation, mention this split in the `emacs-plus` formula and prepare bottles.

daviderestivo:
> Hooray! I fixed the build. Next steps are to update documentation, mention this split in the `emacs-plus` formula and prepare bottles.

Great ... for the bottles I currently use this: https://github.com/daviderestivo/travis-emacs-head-bottle ... feel free to clone it and adapt it to your needs.

BTW I think I will follow your example and split my formula as well since it's become complicated to maintain all of the `if ... else ... end` . statements.

daviderestivo:
> Another valid question - how to 'migrate' users. Should we add a message like 'please use emacs-plus@28`if the user used`--HEAD` option?

I think and if `build.head?` ... with an `odie` and a clear message is the only solution.

d12frosted:
> Great ... for the bottles I currently use this: https://github.com/daviderestivo/travis-emacs-head-bottle ... feel free to clone it and adapt it to your needs.

Cool, thanks!

> BTW I think I will follow your example and split my formula as well since it's become complicated to maintain all of the if ... else ... end . statements.

I can send you a PR 😸 

> I think and if build.head? ... with an odie and a clear message is the only solution.

I agree.

d12frosted:
```
$ brew install emacs-plus --HEAD
Error: 
Emacs 27 and Emacs 28 are now separate formulas. Please use
emacs-plus@27 or emacs-plus@28.

$ brew install emacs-plus@27 [options]

or

$ brew install emacs-plus@28 [options]
```

d12frosted:
I think tomorrow I will merge it 😸

daviderestivo:
> I can send you a PR 😸

Yes please ... so I save some time 😸

daviderestivo:
> $ brew install emacs-plus --HEAD
> Error: 
> Emacs 27 and Emacs 28 are now separate formulas. Please use
> emacs-plus@27 or emacs-plus@28.
> 
> $ brew install emacs-plus@27 [options]
> 
> or
> 
> $ brew install emacs-plus@28 [options]

Yup ... exactly. I like it ...

d12frosted:
Ok, everything is ready. Once it's green, I will merge the PR.

And will build bottles afterwards.

d12frosted:
Here it comes!

PR #227, opened 2020-06-07
[ci] add bottling

d12frosted:
not sure that it will even work 🤔 

plus, I fear that latest macos is the worst thing for Emacs 26 bottle


d12frosted:
BTW, this will not pass due to the #195. Right now GitHub provides only the latest version of macos.

d12frosted:
the last commit is just for fun, to see how CI goes

PR #233, opened 2020-06-13
fix unexec bug in Emacs 26

d12frosted:
based on #231


d12frosted:
Cherry-picked respective commits.

PR #237, opened 2020-06-13
remove HEAD only restriction for with-xwidgets.

beyondpie:


d12frosted:
Thank you! Cherry-picked to master - 41dd588!

Issue #239, opened 2020-06-20
Could not install with --with-modern-icon-cg433n


fladd:
Please make sure to follow these steps (and mark the checkboxes):

- [ x] run `brew update` and try to reproduce the issue again
- [x ] run `brew doctor`, fix all issues and try to reproduce your issue again
- [ x] run `brew config` and `brew doctor` and include their output

## What you were trying to do

I was trying to install with `brew install emacs-plus --with-modern-icon-cg433n`

## What happened (include command output)

<pre>
$ brew install emacs-plus -with-modern-icon-cg433n
</pre>

<details>
  <summary>Command output</summary>
  <pre>

Updating Homebrew...
Usage: brew install [options] formula

Install formula. Additional options specific to formula may be appended to
the command.

Unless HOMEBREW_NO_INSTALL_CLEANUP is set, brew cleanup will then be run for
the installed formulae or, every 30 days, for all formulae.

    -d, --debug                      If brewing fails, open an interactive
                                     debugging session with access to IRB or a
                                     shell inside the temporary build directory.
        --env                        If std is passed, use the standard build
                                     environment instead of superenv. If super
                                     is passed, use superenv even if the formula
                                     specifies the standard build environment.
        --ignore-dependencies        An unsupported Homebrew development flag to
                                     skip installing any dependencies of any
                                     kind. If the dependencies are not already
                                     present, the formula will have issues. If
                                     you're not developing Homebrew, consider
                                     adjusting your PATH rather than using this
                                     flag.
        --only-dependencies          Install the dependencies with specified
                                     options but do not install the formula
                                     itself.
        --cc                         Attempt to compile using the specified
                                     compiler, which should be the name of the
                                     compiler's executable, e.g. gcc-7 for GCC
                                     7. In order to use LLVM's clang, specify
                                     llvm_clang. To use the Apple-provided
                                     clang, specify clang. This option will
                                     only accept compilers that are provided by
                                     Homebrew or bundled with macOS. Please do
                                     not file issues if you encounter errors
                                     while using this option.
    -s, --build-from-source          Compile formula from source even if a
                                     bottle is provided. Dependencies will still
                                     be installed from bottles if they are
                                     available.
        --force-bottle               Install from a bottle if it exists for the
                                     current or newest version of macOS, even if
                                     it would not normally be used for
                                     installation.
        --include-test               Install testing dependencies required to
                                     run brew test formula.
        --devel                      If formula defines it, install the
                                     development version.
[...]

PR #246, opened 2020-07-03
unify icons management

d12frosted:
1. Move all the icons to this repository.
2. Standardise option naming.
   1. No more word duplication (e.g.
      `with-emacs-icons-project-emacs-card-blue-deep`)
   2. No more that long options (see previous point).
   3. Each of the option ends with word icon.
3. Extract option and resource registration to parent class :O

Unfortunately, some of the options are renamed.


d12frosted:
Not sure if it's a good idea to add all icons to the build matrix. But at least I want to make sure that they work lol.

d12frosted:
Btw, also fixes #239

d12frosted:
Ok, I though about not downloading ALL icons unconditionally, but only specified one, but it's better not to do it, which means that resources are tested without additional job. So... removing that configuration.

PR #250, opened 2020-08-11
use emacs-plus@27 by default

d12frosted:


belak:
The documentation in the README should probably be updated to note the new defaults and specify that emacs-plus@26 should be used to install Emacs 26.

d12frosted:
You are right 😸 I have updated readme. And also added CI configuration for testing aliases.

d12frosted:
Merged.

PR #257, opened 2020-08-14
Clean up caveats now that emacs-plus@27 is the default

belak:
This moves the note for other Emacs versions to the emacs-plus@27 formula because it's now the default. It also removes the warning if you use --head with emacs-plus@26 as this seemed old and potentially not needed any more.


d12frosted:
Makes sense! Thank you @belak ! Merged to master.

PR #268, opened 2020-10-05
use workdir for ci actions

d12frosted:


d12frosted:
Emacs 28 doesn't compile, but it is not related to this PR. Master is also red.

PR #270, opened 2020-11-08
Update system appearance patches

ngquerol:
👋 

I just updated these patches so that they work a little bit better.
Now theme changes are applied immediately even when Emacs is not the active application, and there is a new `ns-system-appearance` variable that is set to the current system appearance.

### DIFF
--- a/Formula/emacs-plus@27.rb
+++ b/Formula/emacs-plus@27.rb
@@ -100,7 +100,7 @@ class EmacsPlusAT27 < EmacsBase
 
   patch do
     url (UrlResolver.patch_url "emacs-27/system-appearance")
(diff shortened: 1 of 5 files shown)


d12frosted (approved):
Wonderful!

d12frosted:
Merged into master - 995548b. Thank you very much for this update @ngquerol 💯

d12frosted:
Updated here as well: https://github.com/d12frosted/emacs-plus-basis/commit/043ff79c2fae9c13026fa55acb160f48e853aacc

daviderestivo:
Hi @ngquerol,

thanks for the nice work. I pulled it into https://github.com/daviderestivo/homebrew-emacs-head as well.

Regards,
Davide

PR #278, opened 2020-11-28
[ci] Add macOS versions to build matrix

jidicula:
**Why this change was necessary**
The GitHub workflows were only testing against `macos-latest`, which
still points to macOS 10.15 Catalina[1]. Now that M1 Macs running
macOS 11.0 Big Sur have reached consumers, Emacs Plus should also
include that OS (and CPU arch) in its CI tests.

**What this change does**
 - Adds macOS 11.0 as an option to the build matrix for every
 workflow.
 - Explicitly names macOS 10.15 (Catalina) as an option to the build
 matrix.
 - Explicitly sets Xcode 12.2 if the OS version is 11.0, since
 previous XCode versions cannot build for Big Sur[2].

**Any side-effects?**
None.

**Additional context/notes/links**
[1] https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources
[2] actions/virtual-environments#2056

### DIFF
--- a/.github/workflows/emacs-26.yml
+++ b/.github/workflows/emacs-26.yml
@@ -20,9 +20,10 @@ on:
 
 jobs:
   build:
-    runs-on: macos-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        os: [macos-10.15, macos-11.0]
         build_opts:
           - ""
           - "--build-from-source"
@@ -34,6 +35,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v1
+
+      - name: Use XCode 12.2 for Big Sur
+        if: contains(matrix.os, 'macos-11.0')
+        run: sudo xcode-select -s "/Applications/Xcode_12.2.app"
 
       - name: Build emacs-plus@26 ${{ matrix.build_opts }}
         run: brew install emacs-plus@26.rb ${{ matrix.build_opts }}
--- a/.github/workflows/emacs-27.yml
+++ b/.github/workflows/emacs-27.yml
@@ -20,10 +20,11 @@ on:
 
 jobs:
   build:
-    runs-on: macos-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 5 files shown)


jidicula:
(sorry about the force-push, didn't realize the `contains()` requires single-quoted arguments 😅

jidicula:
@d12frosted I wonder if it would be useful to also include badges for each macOS version so users will know about OS support at a glance from the README. Unfortunately, [matrix builds can't have separate badges yet](https://github.community/t/separate-workflow-badges-when-using-matrix-testing-possible/16708/10), so it might involve breaking up the workflows into separate YAMLs for both OS versions.

d12frosted:
Cool, I like it. I remember that at some point of time `emacs-plus` was broken for specific version of macOS, so having ability to test against few major versions sounds nice. Thank you for contribution!

> sorry about the force-push

No worries regarding force pushes 😸 I will simply cherry-pick commit :) If it would be two commits, I would simply squash them and cherry pick result.

> I wonder if it would be useful to also include badges for each macOS version so users will know about OS support at a glance from the README. Unfortunately, matrix builds can't have separate badges yet, 

Yes, it would be nice. I also had an idea to show each option combination in the README, but as you say... matrix is not supported by badges. So that's fine as is.

> so it might involve breaking up the workflows into separate YAMLs for both OS versions.

I think this is too much. So let's leave it as is for now. Unless it's possible to have a workflow template that would be reused by per-OS workflows.

---

Going to cherry pick it now anyways as I like how it looks now.

d12frosted:
Cherry picked to master (315f55c). Thank you very much for your contribution 😸

jidicula:
> I think this is too much. So let's leave it as is for now. Unless it's possible to have a workflow template that would be reused by per-OS workflows.

I think there might be (see [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses))... I'll do some tryouts with this on my dotfiles repo - if it works, I'll submit another PR here 😄

PR #280, opened 2020-11-29
[ci] add --with-x11 to emacs-plus@28 build matrix

d12frosted:


d12frosted:
The culprit is `cairo`:

```
../src/xterm.h:71:10: fatal error: 'cairo-xlib.h' file not found
#include <cairo-xlib.h>
```

Brew doesn't ship `cairo` with X11 [anymore][1]. I wonder why `pkg-config` doesn't fail and allows me to get this deep into compilation. 🤔 But it's not a big deal.

So one of the options is to... make a copy of `cairo` package. 😭 But I don't really want to maintain a copy. Since I don't use nor cairo nor X11 build, I will constantly forget to update it. But I can create a reminder in `org-mode` to sync versions every once in a while. 🤷 

It all depends on the answer for a simple question - does anyone use `emacs-plus` compiled `--with-x11`? Short search on GitHub shows that some people do use, but they use outdated setups (for example, in their `Brewfile` I see `cairo --with-x11` which is obsolete for some time). 

CC @syl20bnr (I know you was using it, but not sure if it still the case). 

  [1]: https://github.com/Homebrew/homebrew-core/pull/27484

d12frosted:
Resolved conflicts.

syl20bnr:
Hi @d12frosted, I don't use Emacs with X11 anymore. I used to do it when I transitioned from Linux to OSX and was still using i3wm. This is not the case anymore.

Maybe it is OK to stop support of `--with-x11` for Emacs 28.

d12frosted:
@syl20bnr Thanks for the update. Meanwhile even Emacs 27 `--with-x11` is broken 😿

d12frosted:
Fixed on master already

Issue #284, opened 2020-12-12
Pending: `librsvg` option does not work for Apple Silicon (M1) due to Rust dependency


dive:
**Note**: It is not really an issue. I am creating it to track the progress about the current state of Emacs Plus package on the latest Apple Silicon (M1) devices.

## Homebrew state

Homebrew **does not** fully support Apple Silicon at the moment. They recommend using Rosetta 2 as a workaround.

- [macOS 11 Big Sur compatibility on Apple Silicon #7857](https://github.com/Homebrew/brew/issues/7857)
- [Homebrew installation on ARM MacBook with macOS 11.0](https://github.com/Homebrew/discussions/discussions/149)

If you are brave enough, you can try `arm64` Homebrew installation. To make it happen, you have to install Homebrew manually with something like this:

```
sudo rm -Rf /opt/homebrew || true
sudo mkdir -p /opt/homebrew
sudo chown -R $(whoami):staff /opt/homebrew
cd /opt
curl -L https://github.com/Homebrew/brew/tarball/master | tar xz --strip 1 -C homebrew

brew cleanup
brew update
```

Do not forget to add the `/opt/homebrew/bin` to your `$PATH`.

## Emacs Plus state

I am using the following options to install, and I did not try the stable version yet.

```
brew install emacs-plus@28 --with-no-titlebar --without-imagemagick
```

### Dependencies

Note: This is the [current state](https://github.com/Homebrew/brew/issues/7857#issuecomment-741724200) of Homebrew bottles for `arm64`. Also, you can check this issue - [The first ARM (Apple Silicon) bottles are here, help us test them #9488](https://github.com/Homebrew/brew/issues/9488).

| State | Name | Version | Comments |
|:-----:|:-------|:--------|:-----------|
| ✅ | `autoconf` | 2.69 | Bottle |
| ✅ | `gnu-sed` | 4.8 | Source |
| ✅ | `pkg-config` | 0.29.2_3 | Bottle |
| ✅ | `texinfo` | 6.7 | Source |
| ✅ | `gnutls` | 3.6.15 | Source |
| 🛑 | `librsvg` | 2.50.2 | Source, [Failed due to Rust dependency](https://github.com/rust-lang/rust/issues/73908) | 
| ✅ | `little-cms2` | 2-2.11 | Bottle |
| ✅ | `jansson` | 2.13.1 | Source |
| ✅ | `imagemagick` (recommended) | 7.0.10-48 | Source |
[...]

PR #290, opened 2020-12-25
Make librsvg optional for Emacs 27

apriendeau:
If you are a M1 based mac, librsvg will fail to install because it depends on Rust. Rust will not support M1 till 1.49.0 Currently, brew attempts to install 1.48.0.

This PR makes librsvg recommended, giving the option for emacs 27 to be installed on M1.

### DIFF
--- a/Formula/emacs-plus@27.rb
+++ b/Formula/emacs-plus@27.rb
@@ -40,7 +40,7 @@ class EmacsPlusAT27 < EmacsBase
   depends_on "pkg-config" => :build
   depends_on "texinfo" => :build
   depends_on "gnutls"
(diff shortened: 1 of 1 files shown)


dive:
FYI: We have tested the same workaround for Emacs 28 on M1 recently within #284. I was using `emacs-plus` for 3 days in this configuration and did not find any problems.

apriendeau:
Closing because Homebrew fixed support for Rust and Librsvg earlier.

d12frosted:
Thank you for your contribution 💯 Since it's already supported on M1, I will revert my commit against emacs-plus@28.

PR #287, opened 2020-12-19
Fix: Remove homebrew's deprecated x11

benmezger:
This removes the deprecated homebrew's X11 dependency.

Closes issue #286

PR #291, opened 2020-12-25
Fix: Remove homebrew's deprecated x11

wufniks:
This is basically @benmezger 's work from #287.
I only changed dependency from libx11 to libxaw to avoid an error in [this comment](https://github.com/d12frosted/homebrew-emacs-plus/pull/287#issuecomment-748648280).


d12frosted:
Thank you! Cherry-picked each commit to master 💯

benmezger:
Thanks for taking care of this while I was away!

PR #295, opened 2020-12-28
enable NATIVE_FULL_AOT

d12frosted:
I think to make it unconditional. AFAIK, aside from having longer build time (and CI will help me to see real numbers) I has no other drawbacks. 

Examples:

- [Without full AOT][1] - 24 minutes 42 seconds
- [With full AOT][2] - 49 minutes 15 seconds

  [1]: https://github.com/d12frosted/homebrew-emacs-plus/runs/1617715151?check_suite_focus=true
  [2]: https://github.com/d12frosted/homebrew-emacs-plus/runs/1618405833?check_suite_focus=true

### DIFF
--- a/Formula/emacs-plus@28.rb
+++ b/Formula/emacs-plus@28.rb
@@ -104,6 +104,7 @@ class EmacsPlusAT28 < EmacsBase
   end
 
   def install
+    make_flags = []
     args = %W[
       --disable-dependency-tracking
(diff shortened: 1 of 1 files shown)


d12frosted:
Oh well, difference is almost 2x 🤔 does not sound good to me. I will try using `-j$(nproc)`.

But will do it later.

Issue #187, opened 2020-02-23
native-comp branch option


seagle0128:
The native compiling is a amazing feature to improve the Emacs performance. I'd like get an option to build this branch. Thanks!

https://github.com/emacs-mirror/emacs/tree/feature/native-comp

d12frosted:
@seagle0128 hey, do you use any other options? I just need to check if other patches work well together with `native-comp` branch.

seagle0128:
@d12frosted I am using `--with-cocoa`, `--with-jansson` and `--with-xwidgets`. For example,

```shell
brew install emacs-plus --HEAD --with-jansson --without-spacemacs-icon --without-imagemagick@7 --with-cocoa --with-xwidgets
```

d12frosted:
@seagle0128 thanks. I will check if it builds with these options. I can't test it locally, so has to run on CI.

seagle0128:
Thanks  a lot, @d12frosted .

d12frosted:
Don't really understand the error:

```
make[2]: *** [emacs-lisp/autoload.elc] Bus error: 10
make[2]: *** Waiting for unfinished jobs....
make[2]: *** [emacs-lisp/macroexp.elc] Bus error: 10
make[2]: *** [emacs-lisp/byte-opt.elc] Bus error: 10
make[2]: *** [emacs-lisp/cconv.elc] Bus error: 10
make[2]: *** [emacs-lisp/bytecomp.elc] Bus error: 10
make[1]: *** [bootstrap-emacs.pdmp] Error 2
make: *** [src] Error 2
```

Apparently, this branch requires something special to build. I will take a look once I have more time and ability to build locally.

d12frosted:
Quick duckduckgoing leads me to https://akrl.sdf.org/gccemacs.html

Apparently, `libgccjit` must be installed (bundled with `gcc`?).

seagle0128:
Bad news. I found this thread: https://lists.gnu.org/archive/html/emacs-devel/2019-08/msg00039.html.

d12frosted:
@seagle0128 oh well.

atanasj:
Is this feature likely to be introduced?

AndreaCorallo:
Hi, I know zero about Mac and related but maybe this can help?

https://gist.github.com/mikroskeem/0a5c909c1880408adf732ceba6d3f9ab

seagle0128:
[...]

PR #302, opened 2021-01-07
Update README.org

claford-v-lawrence:
There is no need to build jit from source

### DIFF
--- a/README.org
+++ b/README.org
@@ -217,11 +217,7 @@ Please see official [[https://akrl.sdf.org/gccemacs.html][→ gccemacs documenta
 Knows issues:
 
 - =ld: library not found for -lSystem=. 
-  - =$ brew reinstall gcc libgccjit=
-  - =$ sudo ln -s /Library/Developer/CommandLineTools/SDKs/MacOSX11.1.sdk /Library/Developer/CommandLineTools/SDKs/MacOSX11.0.sdk=
-  - =$ sudo ln -s /Library/Developer/CommandLineTools/SDKs/MacOSX11.1.sdk /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk=
-  - =$ export LIBRARY_PATH = "$LIBRARY_PATH:/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib"=
-  - =$ brew reinstall gcc libgccjit=
+  This only happens on older versions of gcc installed by homebrew. Please execute =$ brew reinstall gcc libgccjit= to resolve this issue
 
 
 - Errors during compilation of your =init.el=. Try running Emacs with =-Q=
--- a/README.org
+++ b/README.org
@@ -217,9 +217,13 @@ Please see official [[https://akrl.sdf.org/gccemacs.html][→ gccemacs documenta
 Knows issues:
 
 - =ld: library not found for -lSystem=. 
-  - =$ brew uninstall gcc libgccjit=
-  - =$ brew install gcc=
-  - =$ brew install libgccjit=
+  - =$ brew reinstall gcc libgccjit=
+  - =$ sudo ln -s /Library/Developer/CommandLineTools/SDKs/MacOSX11.1.sdk /Library/Developer/CommandLineTools/SDKs/MacOSX11.0.sdk=
+  - =$ sudo ln -s /Library/Developer/CommandLineTools/SDKs/MacOSX11.1.sdk /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk=
+  - =$ export LIBRARY_PATH = "$LIBRARY_PATH:/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib"=
+  - =$ brew reinstall gcc libgccjit=
+
+
 - Errors during compilation of your =init.el=. Try running Emacs with =-Q=
   option and give it some time to compile everything (maybe run =M-x= to force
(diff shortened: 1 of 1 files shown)


d12frosted:
🤔 the last time I tried on Big Sur, using bottle leads to issues with `LIBRARY_PATH`, and [once I installed libjit from sources][1], everything started to work.

What OS version do you use? 

  [1]: https://github.com/d12frosted/homebrew-emacs-plus/issues/187#issuecomment-751676092

claford-v-lawrence:
I was using macOS Big Sur at the moment. 

I did some close investigation when I was building gccemacs a few weeks back, the problem with the [stock formula](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/libgccjit.rb) is that home-brew is linking against the wrong SDK version. 
The problem was that the stock libgccjit was pointing the system root to `/Library/Developer/CommandLineTools/SDKs/MacOSX11.0.sdk` and my sdk path is `/Library/Developer/CommandLineTools/SDKs/MacOSX11.1.sdk`. And a lot of problem for people who upgraded from Catalina is that their sys root was pointing to `/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk`. 

So I'm gonna add some instructions on correctly linking system sdk and update LIBRARY_PATH. That was gonna be a lot faster than building gcc from source:P

d12frosted (commented):
Oh my god, you did some nice investigation here. Good work 💯 

Regarding recommendations, I am not sure that meddling with system SDK is something `emacs-plus` should recommend. I see you mention two problems here:

1. `MacOSX.sdk` points to `MacOSX10.15.sdk` after upgrade. This is something that users should fix by installing latest command line tools, right? I did nothing special about `MacOSX.sdk` and it points to the correct version.
2. Homebrew provides (or merely provided) broken `libgccjit` bottle. IMHO This is something that should be fixed on their side. And I need to re-test if the latest bottle is OK. 

So while proposed approach results in faster build, these are dirty hacks 🤔 In my sense, rebuilding *only* `libgccjit` from sources is not that bad and is a safer option. On the other hand, it's all about very experimental branch 🤷 

WDYT? 

P. S. I remember @seanfarley told that setting `LIBRARY_PATH` environment variable is an anti-pattern on macOS, but I don't know why. Sean, could you please elaborate? 😸

claford-v-lawrence:
I think you are absolutely right about not modifying sdk path. However, it's just that libgccjit is really slow to build, and is not really different from stock build. I'd love to hear seanfarley's idea:P

seanfarley:
> the problem with the [stock formula](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/libgccjit.rb) is that home-brew is linking against the wrong SDK version.

I mean, this is the crux of the matter. If homebrew's bottle is broken, then why make users jump through even more (fragile) hoops to fix their bug?

I learned about the fragility of `LIBRARY_PATH` in my decade as a MacPorts maintainer. There are many issues with setting this variable. One is that not every process inherits the user's environment variables, e.g. launching Emacs from the spotlight bar won't pick it up. So, now there's *another* hurdle / point of failure: the user must set the environment variable in launchtctl or hack around it via lisp.

Another issue is linking contamination. Setting `LIBRARY_PATH` means that the path is searched *every time* the user compiles / links. Very frequently, this will have unintended consequences, e.g. the compiler might pick up the wrong library to link with. This is *extremely* confusing behavior when it happens.

Finally, and perhaps most importantly, it causes a huge maintenance burden. When `gccemacs` "doesn't work" maintainers now need to go back-and-forth with the user: did you set the environment variable correctly? what about the custom lisp to make sure spotlight picks it up? Even your instructions in this PR don't mention the work needed to be done for picking up `LIBRARY_PATH` in other processes. The user could launch via spotlight, or as a service, or even something else entirely. Perhaps the user installed an elisp package that sources the user's environment variables and everything just works. Or perhaps they didn't and are confused why the instrutcions work for some people and not for them.

All of that headache to work around a homebrew bug? I'd rather the effort be spent in working with the homebrew team to fix the bottle (there are instructions from me for a minimal working example of the bug in the PR). That's the correct solution; and recompiling it locally is the *actual* fix to correct the link path (and works for everyone independent of how they launch emacs or what emacs packages they have installed).

Recompiling `libgccjit` sucks but, ultimately, is the right thing to do. Saving time (for you, the user) is creating more burden for the maintainers of this package.

claford-v-lawrence:
Yeah, I think you're right:P Thanks for all the knowledge, I learned a great deal today from the perspective of a maintainer point of view!

d12frosted:
@seanfarley thank you very much for such detailed response 💯

seanfarley:
FYI: as of today (Jan. 13th, 2021), I just uninstalled and reinstalled both `gcc` and `libgccjit` from bottles and they seem to work; so it seems Homebrew fixed the linking upstream!

d12frosted:
@seanfarley thanks for the headsup. @claford-v-lawrence I would be happy if you could send a PR 😸 I think the section should be still present, but solution can be simplified in addition to not mentioning building from source anymore 💯

claford-v-lawrence:
@d12frosted Done! Thanks for your patience and devotion:P

d12frosted (approved):

d12frosted:
@claford-v-lawrence thank you for your contribution. 💯  Merged as 029113f (with minor style fix).

And @seanfarley thanks for guidance and notifying that now there is no need to install from sources.

PR #324, opened 2021-03-23
Change post-install message for linking to Application

thegeorgeous:
Aliased applications don't appear in applications lists created by apps like Spotlight and Alfred. Default cask behaviour is to copy the `.app` folder to `/Applications`

### DIFF
--- a/Formula/emacs-plus@26.rb
+++ b/Formula/emacs-plus@26.rb
@@ -112,7 +112,7 @@ class EmacsPlusAT26 < EmacsBase
         #{prefix}
 
       To link the application to default Homebrew App location:
-        ln -s #{prefix}/Emacs.app /Applications
+        cp -r #{prefix}/Emacs.app /Applications
     EOS
   end
 
--- a/Formula/emacs-plus@27.rb
+++ b/Formula/emacs-plus@27.rb
@@ -209,7 +209,7 @@ class EmacsPlusAT27 < EmacsBase
         #{prefix}
 
       To link the application to default Homebrew App location:
-        ln -s #{prefix}/Emacs.app /Applications
+        cp -r #{prefix}/Emacs.app /Applications
 
       If you wish to install Emacs 26 or Emacs 28, use emacs-plus@26 or
       emacs-plus@28 formula respectively.
--- a/Formula/emacs-plus@28.rb
+++ b/Formula/emacs-plus@28.rb
@@ -250,7 +250,7 @@ class EmacsPlusAT28 < EmacsBase
         #{prefix}
 
       To link the application to default Homebrew App location:
-        ln -s #{prefix}/Emacs.app /Applications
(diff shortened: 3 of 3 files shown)


d12frosted:
Nice, thanks. 

BTW, for some reason spotlight finds my Emacs.app without `ln -s` or copying 🤔 Can't understand how. I am usually starting Emacs with a key chord.

P. S. ignore failed tests, `--with-x11` is broken right now.

thegeorgeous:
@d12frosted, I didn't have a problem with links before either. I recently switched to a new Macbook that has M1 and Big Sur and both Spotlight and Alfred were not listing Emacs and it was the only application using symlinks. Removing the link and copying the `.app` file fixed it. Not sure what changed, but in any case, I think copying the `.app` file should be the standard behaviour.

jmhammond:
FYI @thegeorgeous, since `Emacs.app` is a directory, you need to specify `cp -r` for recursive copying to the `/Applications` directory.

d12frosted:
What I like about linking is that upgrading/reinstalling of Emacs doesn't require performing any extra steps with `Emacs.app` in Applications folder. When you copy, you need to copy after each upgrade/reinstall. Maybe the formula should simply copy `Emacs.app`, so that users do not need to perform any extra steps?

jmhammond:
> What I like about linking is that upgrading/reinstalling of Emacs doesn't require performing any extra steps with `Emacs.app` in Applications folder. When you copy, you need to copy after each upgrade/reinstall. Maybe the formula should simply copy `Emacs.app`, so that users do not need to perform any extra steps?

That would be huge for new users like me. The is my first Mac and I was confused when I had to do extra steps to get `emacs-mac` to be picked up by Spotlight.

d12frosted:
I tried placing the following code in the end of `if (build.with? "cocoa") && (build.without? "x11")` block, but it doesn't work. I mean, nothig happens. When I try to put `Emacs.app` into `/Applications`, I get an access error (which is expected), so I wanted to put them into user applications folder, but aparently brew doesn't allow it.

```ruby
+      apps_dir = File.expand_path "~/Applications"
+      mkdir_p apps_dir
+      rm_rf "#{apps_dir}/Emacs.app"
+      cp_r "#{prefix}/Emacs.app", "#{apps_dir}/Emacs.app"
```

wildwestrom:
Hey guys, check #345. This is why `build (macos-11.0, --with-x11)` failed.

webframp:
This change it will probably make for a simpler experience since links aren't found by default. In my case, I added /usr/local/Cellar to my search path for Alfred and it works:
﻿
image

marcus233:
You can tell Alfred to include alias files.

Alfred Settings -> General -> Default Results -> Advanced 
Add `com.apple.alias-file`

d12frosted:
Fixed by #642

PR #352, opened 2021-05-06
Add option to not compress some files (.el, .info, .man, etc.)

mnewt:
This is a simple PR to mirror the `configure` option `--without-compresss-install`.

The main reason I find this valuable is it makes doing searches over the `.el` files easier and quicker.

I've tested it locally and seems to work fine. Please let me know what you think.

### DIFF
--- a/Formula/emacs-plus@28.rb
+++ b/Formula/emacs-plus@28.rb
@@ -10,6 +10,7 @@ class EmacsPlusAT28 < EmacsBase
 
   # Opt-out
   option "without-cocoa", "Build a non-Cocoa version of Emacs"
+  option "without-compress-install", "Don't compress some files (.el, .info, etc.)"
 
   # Opt-in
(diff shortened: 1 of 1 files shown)


d12frosted:
Hey,

In general I try to keep amount of options as low as possible. And if possible, enable/disable something uncoditionally. Could you please elaborate a little bit more on the following statement?

> The main reason I find this valuable is it makes doing searches over the .el files easier and quicker.

Do you mean full-text search in the lisp directory? Would be nice to understand why they are compressed by default 😅 Gonna do my homework haha.

mnewt:
No worries at all if you don’t want to add it. Keeping things simple is a good goal. 

I believe the reason those things are normally compressed as part of the build process is to save disk space for the resulting installation. 

Yes, I regularly do full text searches of the lisp directory. I guess that’s unusual? :)

PR #365, opened 2021-06-30
(ci) update Xcode to 12.5

mkvoya:
All the [workflows](https://github.com/d12frosted/homebrew-emacs-plus/actions) have been failed for a few days. After some investigation, some actions may be required.

- [Support for Xcode 12.2 has been removed](https://github.com/actions/virtual-environments/issues/3555). Thus we'd better switch to a later version such as Xcode 12.4 or 12.5.
- The yaml tag is `macos-11` rather than `macos-11.0`, I don't know whether `macos-11.0` still works.

I have NOT tested the changes by myself since the `macos-11` has been [moved to private pools](https://github.com/actions/virtual-environments/issues/2486) and my fork requires some kinds of special applications to use `macos-11`.

### DIFF
--- a/.github/workflows/emacs-26.yml
+++ b/.github/workflows/emacs-26.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-11.0]
+        os: [macos-11]
         build_opts:
           - ""
           - "--build-from-source"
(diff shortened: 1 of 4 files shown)


d12frosted:
Hey @mkvoya 

CI is indeed red for some time, hadn't time to dig in. Thanks for investigation and your PR. Let's see if it works. In the worst case let's switch to older macos.

d12frosted:
Thank you! 💯 Merged as 5a2c9c6. Native comp is broken due to upstream issue.

Issue #364, opened 2021-06-27
Installs ok - strange function error on first run


tigger04:
Please make sure to follow these steps (and mark the checkboxes):

- [x] run `brew update` and try to reproduce the issue again
- [x] run `brew doctor`, fix all issues and try to reproduce your issue again
- [x] run `brew config` and `brew doctor` and include their output

## What you were trying to do

Install emacs-plus@28 on to a brand new macbook with:

`brew install emacs-plus@28 --with-modern-paper-icon --with-xwidgets --with-native-comp --with-no-titlebar --with-mailutils --with-debug
`

## What happened (include command output)

Brew reports it installed correctly

```
* brew info emacs-plus@28
d12frosted/emacs-plus/emacs-plus@28: stable 28.0.50
/usr/local/Cellar/emacs-plus@28/28.0.50 (4,330 files, 167MB) *
  Built from source on 2021-06-27 at 15:31:24 with: --with-no-titlebar --with-debug --with-xwidgets --with-native-comp --with-mailutils --with-modern-paper-icon
From: https://github.com/d12frosted/homebrew-emacs-plus/blob/HEAD/
==> Dependencies
Build: autoconf ✔, gnu-sed ✔, pkg-config ✔, texinfo ✔
Required: gnutls ✔, librsvg ✔, little-cms2 ✔, jansson ✔
Recommended: imagemagick ✔
Optional: dbus ✔, mailutils ✔
==> Options
--with-EmacsIcon1-icon
	Using Emacs EmacsIcon1 icon
--with-EmacsIcon2-icon
	Using Emacs EmacsIcon2 icon
--with-EmacsIcon3-icon
	Using Emacs EmacsIcon3 icon
--with-EmacsIcon4-icon
	Using Emacs EmacsIcon4 icon
--with-EmacsIcon5-icon
	Using Emacs EmacsIcon5 icon
--with-EmacsIcon6-icon
	Using Emacs EmacsIcon6 icon
--with-EmacsIcon7-icon
	Using Emacs EmacsIcon7 icon
--with-EmacsIcon8-icon
	Using Emacs EmacsIcon8 icon
--with-EmacsIcon9-icon
	Using Emacs EmacsIcon9 icon
--with-cacodemon-icon
	Using Emacs cacodemon icon
--with-ctags
	Don't remove the ctags executable that Emacs provides
--with-dbus
	Build with dbus support
--with-debug
	Build with debug symbols and debugger friendly optimizations
[...]

PR #366, opened 2021-06-30
Pin the emacs 28 formula to the commit before builds started failing

bienjensu:
This PR would pin the emacs-plus@28 formula to the commit before emacs-mirror/emacs@5dd2d50 to temporarily fix an issue occurring upstream with native-comp builds. Decided to update the main formula, but possibly could add another formula, maybe `emacs-plus@28-pin` or something and remove native comp from the original formula and explain why in the readme. See discussion under #364.

### DIFF
--- a/Formula/emacs-plus@28.rb
+++ b/Formula/emacs-plus@28.rb
@@ -64,7 +64,7 @@ class EmacsPlusAT28 < EmacsBase
   # URL
   #
 
-  url "https://github.com/emacs-mirror/emacs.git"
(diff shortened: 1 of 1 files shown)


d12frosted (approved):
Thanks. It makes sense to do it in the main formula (as I usually did before). Let's just revert once the upstream issue is solved.

P. S. sorry for not making it by myself sooner.

d12frosted:
Merged as d94ede3.

PR #367, opened 2021-07-01
Add patch to fix native-comp MacOS self contains app build issue

karlbright:
This fixes an issue that was raised in #364 and a fix provided by @bienjensu in #366. The difference between the two PRs is that this uses an upstream patch on the mailing list, and avoids pinning the commit sha.

I have a feeling this will be a temporary fix and will see a fix upstream later for this. However, as it stands, the builds are broken without either this fix or the pinning fix in #366

> I believe commit 5dd2d50 which moved *.eln files and various paths
> around a bit for macOS builds, has re-introduced an old bug from last
> year. Basically the checksums that makes up part of the *.eln file
> names uses the absolute path of the .el file in question, but for self
> contained .app builds it needs to just use the relative path to the
> app itself.

This applies a patch created and providing by Alan Third via the mailing
list.

Source of this patch can be found on the bug-gnu-emacs mailing list.
https://lists.gnu.org/archive/html/bug-gnu-emacs/2021-06/msg01381.html


karlbright:
Closing in favor of already merged #366

Issue #276, opened 2020-11-16
Support HARFBUZZ and Cairo for ligature


mikelxc:
Emacs-plus doesn't come with ligature support. 
While searching for enabling methods, I've found out that Mac port of Emacs (by Mitsuharu Yamamoto) for macOS has this enabled by default with (mac-auto-operator-composition-mode)
And there's a [ligature.el](https://github.com/mickeynp/ligature.el) which requires HARFBUZZ, and cairo. Both of them are not supported by emacs-plus@27 by default.

mikelxc:
related issue #222

mikelxc:
Work arounds, 
Add
`(when (window-system)
  (set-frame-font "Fira Code"))
(let ((alist '((33 . ".\\(?:\\(?:==\\|!!\\)\\|[!=]\\)")
               (35 . ".\\(?:###\\|##\\|_(\\|[#(?[_{]\\)")
               (36 . ".\\(?:>\\)")
               (37 . ".\\(?:\\(?:%%\\)\\|%\\)")
               (38 . ".\\(?:\\(?:&&\\)\\|&\\)")
               (42 . ".\\(?:\\(?:\\*\\*/\\)\\|\\(?:\\*[*/]\\)\\|[*/>]\\)")
               (43 . ".\\(?:\\(?:\\+\\+\\)\\|[+>]\\)")
               (45 . ".\\(?:\\(?:-[>-]\\|<<\\|>>\\)\\|[<>}~-]\\)")
               (46 . ".\\(?:\\(?:\\.[.<]\\)\\|[.=-]\\)")
               (47 . ".\\(?:\\(?:\\*\\*\\|//\\|==\\)\\|[*/=>]\\)")
               (48 . ".\\(?:x[a-zA-Z]\\)")
               (58 . ".\\(?:::\\|[:=]\\)")
               (59 . ".\\(?:;;\\|;\\)")
               (60 . ".\\(?:\\(?:!--\\)\\|\\(?:~~\\|->\\|\\$>\\|\\*>\\|\\+>\\|--\\|<[<=-]\\|=[<=>]\\||>\\)\\|[*$+~/<=>|-]\\)")
               (61 . ".\\(?:\\(?:/=\\|:=\\|<<\\|=[=>]\\|>>\\)\\|[<=>~]\\)")
               (62 . ".\\(?:\\(?:=>\\|>[=>-]\\)\\|[=>-]\\)")
[...]

PR #371, opened 2021-07-14
Add modern fonts

elken:
Includes harfbuzz & cairo for modern font rendering/better ligature
support by default, can be disabled if required.

Ref: #276

### DIFF
--- a/Formula/emacs-plus@28.rb
+++ b/Formula/emacs-plus@28.rb
@@ -10,6 +10,7 @@ class EmacsPlusAT28 < EmacsBase
 
   # Opt-out
   option "without-cocoa", "Build a non-Cocoa version of Emacs"
+  option "without-modern-fonts", "Disable modern font rendering (not recommended)"
 
   # Opt-in
   option "with-ctags", "Don't remove the ctags executable that Emacs provides"
@@ -35,6 +36,12 @@ class EmacsPlusAT28 < EmacsBase
   depends_on "imagemagick" => :recommended
   depends_on "dbus" => :optional
   depends_on "mailutils" => :optional
+
+  if build.with? "modern-fonts"
+    depends_on "freetype"
+    depends_on "cairo"
+    depends_on "harfbuzz"
+  end
 
   if build.with? "x11"
     depends_on "libxaw"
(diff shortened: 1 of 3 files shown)


Formula/emacs-plus@28.rb
```diff
@@ -10,6 +10,7 @@ class EmacsPlusAT28 < EmacsBase
 
   # Opt-out
   option "without-cocoa", "Build a non-Cocoa version of Emacs"
+  option "without-modern-fonts", "Disable modern font rendering (not recommended)"
```
d12frosted, line 13: What is disadvantage of `modern-fornts`? I mean, is it really should be an opt-out option or just an unconditional feature?

elken: None that I'm aware of, but it's still considered new enough that I think it's worth having the option to turn it off.

d12frosted:
Thank you @elken !

elken:
> Thank you @elken !

Thanks for emacs-plus :)

aaronjensen:
Is there a reason that upstream would not accept this patch?

elken:
Haven't gotten around to submitting yet.

aaronjensen:
Okay, thanks. I just wanted to be aware if there was some upstream concern about it. Seems like possibly not. I'll give it a shot today. Thanks!

elken:
There's almost definitely a better way of doing it so it'll likely get rejected, but the point of the patch is there's no way for macos to build in harfbuzz as it depends on x11.

aaronjensen:
In my experience they are helpful and responsive, so I'm sure they could help you do it in a way they'd be good with.

dsdshcym:
Based on this comment from Alan, it seems that Harfbuzz doesn't have any particular advantages over CoreText
https://www.reddit.com/r/emacs/comments/byddvm/emacsdevel_harfbuzz_is_now_available_on_master/eqnkf9o/?context=3
> there's no particular advantage to Harfbuzz over macOS's CoreText, so Yamamoto Mitsuharu, maintainer of the Mac port and the person who did the final Harfbuzz work, didn't see any point in porting it.

And I can enable ligatures on emacs-plus@28 without Harfbuzz
So I doubt there would be any difference 🤔

aaronjensen:
Interesting. @elken did you notice any concrete difference?

elken:
I did a few visual tests and I thought so but I guess I was wrong. Happy for this to be reverted if there's no significant change.

d12frosted:
I also don't see any difference. @elken feel free to send a PR :)

idrozd:
@dsdshcym may I ask how exactly do you enable ligatures?

This?

```elisp
(unicode-fonts :variables unicode-fonts-enable-ligatures t)
```

squiter:
@dsdshcym sorry to revive this, but how did you enable the ligatures without the harfbuzz?

dsdshcym:
It's been a while, so I kind of forget what I was using

FWIW I'm now using doom-emacs's ligature module to enable ligatures,
see here:
https://github.com/hlissner/doom-emacs/blob/1b1556c4766529cb75837d9d478be1d9bc491078/modules/ui/ligatures/config.el#L180-L193

PR #390, opened 2021-09-04
Fix system appearance patch for Emacs 28

### LINKED ISSUE #369: Natural title bar has bad behavior in emacs-28
I installed emacs-28 in my laptop on MacOS Big Sur. But the title bar seems to be different from before:
image
Look at the title bar: the font is dark even on a dark theme, which is not the case for emacs-27.2. And the font is not at the middle of title-bar.

I'm using doom-emacs configuration. I'm not sure whether this is a problem of doom or emacs-plus. But everything is OK on emacs-27.

Any hint about this?

ngquerol:
Closes #369

Essentially, make sure we correctly set the current frame's appearance when updating the `ns-appearance` frame parameter.


d12frosted (approved):
Thank you very much 🙏

PR #399, opened 2021-10-04
Use emacs-28 release branch for emacs-plus@28 build

kommen:
See https://mail.gnu.org/archive/html/emacs-devel/2021-09/msg02290.html

Otherwise emacs-plus@28 will install from master, which is Emacs 29.0.50 now.

### DIFF
--- a/Formula/emacs-plus@28.rb
+++ b/Formula/emacs-plus@28.rb
@@ -64,7 +64,7 @@ class EmacsPlusAT28 < EmacsBase
   # URL
   #
 
-  url "https://github.com/emacs-mirror/emacs.git"
+  url "https://github.com/emacs-mirror/emacs.git", :branch => "emacs-28"
 
   #
   # Icons


kommen:
It appears there is a problem with the emacs-28 branch with `no-titlebar` patch

From the failing CI build:

```
==> Applying no-titlebar.patch

patching file src/nsterm.m
Hunk #2 FAILED at 8293.
1 out of 2 hunks FAILED -- saving rejects to file src/nsterm.m.rej
```

d1egoaz:
Thanks @kommen 

Meanwhile, I've moved to use the emacs28 branch following:

```
brew edit emacs-plus@28
```

and applying your patch:

```diff
-  url "https://github.com/emacs-mirror/emacs.git"
+  url "https://github.com/emacs-mirror/emacs.git", :branch => "emacs-28"
```

d12frosted (approved):
Great! Thanks @kommen :100: 

Yeah, that patch is broken, I can't find time to fix that. Side note: if someone wants to get it quicker - PR is welcome ;)

PR #406, opened 2021-10-14
Add tab-bar support

d1egoaz:
It was recently added to master in https://lists.gnu.org/archive/html/emacs-devel/2021-10/msg01006.html

Tested with http://ruzkuku.com/texts/emacs-global.html kind of global mode line and it worked fine
I installed emacs 28 with ` brew install emacs-plus@28 --with-no-frame-refocus --with-native-comp --with-xwidgets --with-tab-bar`

```lisp
(setq tab-bar-format '(
                        tab-bar-format-tabs-groups
                        tab-bar-format-align-right
                        tab-bar-format-global))
```

### DIFF
--- a/Formula/emacs-plus@28.rb
+++ b/Formula/emacs-plus@28.rb
@@ -19,6 +19,7 @@ class EmacsPlusAT28 < EmacsBase
   option "with-xwidgets", "Experimental: build with xwidgets support"
   option "with-no-frame-refocus", "Disables frame re-focus (ie. closing one frame does not refocus another one)"
   option "with-native-comp", "Build with native compilation"
+  option "with-tab-bar", "Build with tab-bar support"
 
   #
   # Dependencies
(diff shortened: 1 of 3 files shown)


d12frosted:
I see. Since it does nothing unless the respective mode is enabled, it should be included unconditionally to reduce amount of options. And seems like Eli is willing to include this patch to `emacs-28` branch so let's wait a week and see what happens. If it's not getting into `emacs-28`, we can add it as a patch here.

d1egoaz:
tab-bar support was added to emacs-28 branch! 🎉

d12frosted:
> tab-bar support was added to emacs-28 branch! 🎉

I am sorry that you spent time on this PR that wasn't merged. Still, many thanks for your effort 💯

d1egoaz:
> > tab-bar support was added to emacs-28 branch! 🎉
> 
> I am sorry that you spent time on this PR that wasn't merged. Still, many thanks for your effort 💯

No worries, it wasn't too much time either :D , the repo configuration makes easy to test upstream patches. and TBH I prefer to have these features directly from upstream.

Issue #423, opened 2021-11-13
Why is Emacs built with imagemagick by default in this formula?


aaronjensen:
What is the historical (and current) reason that we build with imagemagick here? What does it enable on macOS that we don't get out of the box?

d12frosted:
Honestly, at this point I don't remember anymore. I think back than it was a way to view various images inside of Emacs. And now all I can say about imagemagick - it's a [constant](https://github.com/d12frosted/homebrew-emacs-plus/issues/310) _ [source](https://github.com/d12frosted/homebrew-emacs-plus/issues/11) of [build](https://github.com/d12frosted/homebrew-emacs-plus/issues/44) _ [problems](https://github.com/d12frosted/homebrew-emacs-plus/issues/75) _ [again and again](https://github.com/d12frosted/homebrew-emacs-plus/issues/161). It's been quite for a while, but if enough people think that it's not needed, I am all for removing it at once.

aaronjensen:
I've got a build without it right now and I can open images and manipulate them just fine. Maybe it's worth making it opt-in for a while and seeing what happens

d12frosted:
@aaronjensen I agree. Would you like to contribute that? I think 28 and 29 are good targets for this experiment.

aaronjensen:
I'm going to close this out, but if anyone has problems w/ not having imagemagick installed, please let us know.

PR #425, opened 2021-11-14
imagemagick is not included by default in emacs-28 and emacs-29

aaronjensen:
Re: #423


d12frosted (approved):
🙏 thank you very much

PR #435, opened 2022-02-04
(emacs-29) remove no-titlebar option

d12frosted:
Will merge it in march


d12frosted:
due to war it slipped out of my mind
but time has come to say goodbye to this patch

PR #449, opened 2022-04-09
provide utility for dealing with icons

d12frosted:
The provided script goes over all available icons in the
icons/original folder and does the following things:

1. creates iconset
2. creates preview
3. Icons.rb with name of the icon and its sha256
4. Adds icon to the build matrix

The README file is not updated automatically (yet?).

This basically brings the following benefits:

1. greatly simplifies addition of new icons (because all/most of the
   machinery is done by the script)
2. even further standardises icons provided by emacs-plus

---

Once this build is green, I would be very grateful if someone tests this change on a real machine.

Basically, you just need to switch to `feature/iconset-tool` branch in the `homebrew` tap directory and then install with some non-default icon of your choice. And then verify that icon looks good :)


d12frosted:
Seems like it's working. I am yet to test it on macos machine (hopefully tomorrow I'll get my hands on it). Latest commit simply optimizes CI.

d12frosted:
As a side note - maybe rebuilding ALL icons is not a good idea. On the other hand, configured it not to happen that often, only when icons are changed. Alternative would be to generate workflow per icon :shrug: I will tackle it out of the context of this PR.

d12frosted:
Tested on a mac, works.

Issue #414, opened 2021-11-02
Get corrent environment variable with fish as default shell


404cn:
After zinit repo has been deleted I had switch to fish from zsh and set it as default shell on macOS 12.0.1, but when I run `genenv SHELL` on emacs-plus@28 I get `/bin/zsh`. 

I use emacs to write Golang program, and run Golang's `os.Getenv()` also can't get environment variable which I write this into my `config.fish` with `set -gx foo bar`.

I also install [exec-path-from-shell](https://github.com/purcell/exec-path-from-shell) but I have to set `exec-path-from-shell-shell-name` to `/usr/local/bin/fish` and it can read `PATH` into emacs, but other environment variable much run `exec-path-from-shell-copy-env` manually or Golang's `os.Getenv()` won't get this environment variable which means I can't run my program correctly on emacs-plus@28.

I also test this on [emacsmacport](https://github.com/railwaycat/homebrew-emacsmacport) and it can get correct `SHELL` which is `/usr/local/bin/fish`, and I don't need to set `exec-path-from-shell-name` manually cause it can get correct `SHELL`, and Golang's `os.Getenv()` works too.

d12frosted:
Hey @404cn,

Environment variables is always a burden on macos and is a constant topic of confusion and troubles. Issue tracker of Emacs+, Spacemacs and Doom Emacs are full of issues caused by invalid environment. 

I am not sure how it works in Emacs Mac Port, but they have so many patches for MacOS that it's not GNU Emacs anymore.

I have few questions though:

1. How do you start Emacs? I suspect that via Finder/Spotlight by opening Emacs.app.
2. How do you set your default shell? By using `chsh` or by changing configuration of your terminal emulator, for example Terminal.app? The fact that you get `zsh` tells me that something is fishy (not related to `fish`) happening here.

404cn:
> Hey @404cn,
> 
> Environment variables is always a burden on macos and is a constant topic of confusion and troubles. Issue tracker of Emacs+, Spacemacs and Doom Emacs are full of issues caused by invalid environment.
> 
> I am not sure how it works in Emacs Mac Port, but they have so many patches for MacOS that it's not GNU Emacs anymore.
> 
> I have few questions though:
> 
>     1. How do you start Emacs? I suspect that via Finder/Spotlight by opening Emacs.app.
> 
[...]

Issue #378, opened 2021-08-04
Need to set LIBRARY_PATH for emacs-plus@28 with --with-native-comp


poldy:
I ran `brew install emacs-plus@28 --with-native-comp` and it completed successfully. However, when I tried to run Emacs although the editor worked, there were loads of native compilation warnings.

The solution was to create a one-line *~/.emacs.d/early-init.el*, with the following contents:

```lisp
(setenv "LIBRARY_PATH" "/opt/homebrew/lib/gcc/11:/opt/homebrew/lib/gcc/11/gcc/aarch64-apple-darwin20/11.1.0")
```

I'd suggest adding this to the "gccemacs" section of the README.

This is on an M1 iMac, it works great now.

jilen:
Start `emacs` from the `Terminal` instead of click the emacs icon works for me.

kommen:
I can confirm setting `LIBRARY_PATH` as @poldy suggests as an env in my bash/zsh config fixes native-comp errors. Without it set I got this (a bit redacted)

```
Native compiler error: (lambda (&optional arg6) (let ((f #'kill-emacs)) (funcall f arg6))), Compiling <home>/.emacs.d/eln-cache/28.0.50-b1d099bb/subr--trampoline-6b696c6c2d656d616373_kill_emacs_0.eln...
ld: library not found for -lgcc_s.1.1
libgccjit.so: error: error invoking gcc driver
Debugger entered--Lisp error: (native-ice "failed to compile" "<home>/emacs.d/eln-cache/28.0.50-b1d099bb/..." "error invoking gcc driver")
```

d12frosted:
See this [message][1]. If you have to set `LIBRARY_PATH` that probably just means something is wrong with your environment. Works for me without any hacks like this. Though I must admin, never tested on M1.

[1]: https://github.com/d12frosted/homebrew-emacs-plus/issues/187#issuecomment-751398304

kommen:
@d12frosted thanks!

After reinstalling gcc and libgccjit (also from source) with no success, I poked around in my env:

[...]

PR #453, opened 2022-04-25
emacs-plus@29: inject PATH variable into Info.plist file

d12frosted:
The core issue here is environment in which
macOS starts applications from Finder/Docker/Spotlight/etc. This
change mostly helps users of `native-comp` feature (e.g. #378), but
it also should help other users as well ([see this comment][1])

[1]: https://github.com/d12frosted/homebrew-emacs-plus/issues/414#issuecomment-957166631

### DIFF
--- a/Formula/emacs-plus@29.rb
+++ b/Formula/emacs-plus@29.rb
@@ -3,6 +3,7 @@ require_relative "../Library/EmacsBase"
 class EmacsPlusAT29 < EmacsBase
   init 29
   version "29.0.50"
+  env :std
 
   #
   # Options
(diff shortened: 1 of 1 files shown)


d12frosted:
Will merge it on Wednesday unless someone reports issues.

gildo:
hi @d12frosted, is this patch removing the support to brew service from emacs-29?

d12frosted:
@gildo no, this simply injects `PATH` variable value to `Info.plist` file as the title says. You can read more about this feature in [the separate section of readme file](https://github.com/d12frosted/homebrew-emacs-plus#injected-path). Why do you think that it's related to `brew services`? And most importantly, why do you think I am removing support?

roman-rudakov:
@d12frosted as far as I can see, the `def plist` section is removed, and when I build emacs-plus@29 I don't see `emacs-plus@29` in the `brew services list` anymore.

d12frosted:
@roman-rudakov see https://github.com/d12frosted/homebrew-emacs-plus/pull/467

PR #458, opened 2022-04-28
emacs-28: fix native comp on intel with std env

d12frosted:


d12frosted:
It doesn't make sense as `emacs-plus@28` is not using standard environment.

PR #463, opened 2022-05-10
emacs@28: switch to standard env

d12frosted:
It's the first step towards PATH injection to Emacs.app as it is done
in emacs-plus@29 (see #453).

It has changes from #453, #457, #460 and #462.

Next step would be to actually inject PATH as it is done in #453.


d12frosted:
Ok, checked configure file of Emacs 28 and it seems like it's not supporting brew-installed libgccjit, so have to revert changes to flags manipulations. At least in Emacs 29 it works :)

PR #470, opened 2022-05-19
emacs-{28,29}: improve path injection routine

d12frosted:
1. Move it to a helper function to the EmacsBase class
2. Be more verbose about injection process
3. Use PATH helper provided by brew to simplify manipulations


d12frosted:
I've canceled runs on unaffected Emacs versions.

d12frosted:
IMO it should be better with this change, so merging.

PR #478, opened 2022-07-07
emacs-29: fix compilation of native-comp feature

d12frosted:


d12frosted:
Race condition of checks. It actually passed.

PR #492, opened 2022-08-09
emacs-29: fix libgccjit discovery patch

d12frosted:


d12frosted:
Succesfuly compiled locally.

hkjels:
Just did another try with `brew install emacs-plus@29 --with-xwidgets --with-native-comp --with-modern-sjrmanning-icon
`, but still getting:
```
checking for dlfunc... no
checking for gcc_jit_context_acquire in -lgccjit... no
configure: error: ELisp native compiler was requested, but libgccjit was not found.
```

hkjels:
Running `echo $(dirname $(brew ls -v libgccjit | grep -E 'libgccjit\.(so|dylib)$' | tail -1))` works though.

hkjels:
Works fine without the `--with-native-comp` flag though. I noticed a performance boost on my old intel with that flag, but I haven't really spotted much difference on this M1, so I'm fine with this for now.

d12frosted:
@hkjels make sure to `brew update` before `brew install emacs-plus@29 --with-native-comp`. 

re prefromance: I haven't seen much improvement even on Intel. I mean, it does many things faster. But in most cases I care only about interactive part, which didn't improve much. On the other hand, IMO native compilation is inconvinient - you need to compile a lot of stuff, sometimes it recompiles in the most unpexteced time, sometimes it breaks. 🤷

PR #493, opened 2022-08-10
Add --with-compress-install option (fix Emacs 29 failing to start)

linnik:
This is an issue for the Emacs 29 formula only.

See https://github.com/syl20bnr/spacemacs/issues/11585 and
https://mail.gnu.org/archive/html/emacs-devel/2022-08/msg00234.html
for motivation.

It's common for Emacs configs to have a line such as
(setq file-name-handler-alist nil), which optimizes early loading of packages,
until a further point where the variable is restored. Both Spacemacs and Doom
Emacs do this. However, it means anything other than plain text (e. g. compressed
.el.gz files) cannot be understood by Emacs until the value is reset. It used to
due to a bug where compressed files where treated differently regardless of
file-name-alist, but that's recently been fixed, breaking compressed install
such as nix's with a cryptic (void-variable \233) error. 

This PR disables optimization by default and allows a user to opt in to compression, making the optimization work again.
It should probably be mentioned in the manual or emacs-overlay as well.

### DIFF
--- a/Formula/emacs-plus@29.rb
+++ b/Formula/emacs-plus@29.rb
@@ -19,6 +19,7 @@ class EmacsPlusAT29 < EmacsBase
   option "with-xwidgets", "Experimental: build with xwidgets support"
   option "with-no-frame-refocus", "Disables frame re-focus (ie. closing one frame does not refocus another one)"
   option "with-native-comp", "Build with native compilation"
+  option "with-compress-install", "Build with compressed install optimization"
 
   #
   # Dependencies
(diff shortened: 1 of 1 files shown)


d12frosted (commented):
I like it. Let's try it and see if it helps.

vinurs:
this still need to update readme

d12frosted:
@vinurs I agree with you. Though I want to see how it plays. If there are no downsides of the default behaviour, I would rather just disable compress install unconditionally. If it creates issues, then the default behaviour needs to be reverted and the option needs to be documented :) So I am playing lazy games here 😂

PR #500, opened 2022-09-01
Unlink "man1/ctags.1" file when without ctags and compress-install

Hirozy:
When witout ctags and compress-install, "man1/ctags.1" will be linked instead of "man1/ctags.1.gz", which will conflict with universal-ctags (maybe ctags as well). The error details are as follows

```
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /opt/homebrew
Could not symlink share/man/man1/ctags.1
Target /opt/homebrew/share/man/man1/ctags.1
is a symlink belonging to universal-ctags. You can unlink it:
  brew unlink universal-ctags

To force the link and overwrite all conflicting files:
  brew link --overwrite emacs-plus@29

To list all files that would be deleted:
  brew link --overwrite --dry-run emacs-plus@29

Possible conflicting files are:
/opt/homebrew/share/man/man1/ctags.1 -> /opt/homebrew/Cellar/universal-ctags/p5.9.20220828.0/share/man/man1/ctags.1
```

### DIFF
--- a/Formula/emacs-plus@29.rb
+++ b/Formula/emacs-plus@29.rb
@@ -232,6 +232,8 @@ class EmacsPlusAT29 < EmacsBase
       (bin/"ctags").unlink
       if build.with? "compress-install"
         (man1/"ctags.1.gz").unlink
+      else
+        (man1/"ctags.1").unlink
       end
     end
   end


d12frosted (approved):
Good point! Thanks!

PR #507, opened 2022-09-15
CFLAGS="-DFD_SETSIZE=10000 -DDARWIN_UNLIMITED_SELECT close #487

### LINKED ISSUE #487: how to CFLAGS="-DFD_SETSIZE=10000 -DDARWIN_UNLIMITED_SELECT "?
see https://emacstalk-github-io.translate.goog/post/022/?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=zh-CN

the default value of max open files is 1024 which is too small for me .
I can increment it with  this patch, hope homebrew-emacs-plus can support this .

```
diff --git a/Formula/emacs-plus@29.rb b/Formula/emacs-plus@29.rb
index 87b9f0d..c25bb5c 100644
--- a/Formula/emacs-plus@29.rb
+++ b/Formula/emacs-plus@29.rb
@@ -114,6 +114,7 @@ class EmacsPlusAT29 < EmacsBase
     args << "--with-native-compilation" if build.with? "native-comp"
 
     ENV.append "CFLAGS", "-g -Og" if build.with? "debug"
+    ENV.append "CFLAGS", " -DFD_SETSIZE=10000 -DDARWIN_UNLIMITED_SELECT "
 
     args <<
       if build.with? "dbus"
```

```
export CFLAGS="-DFD_SETSIZE=10000 -DDARWIN_UNLIMITED_SELECT" // this does not work
```

jixiuf:


Formula/emacs-plus@26.rb
```diff
@@ -57,6 +57,7 @@ def install
     imagemagick_lib_path = Formula["imagemagick@6"].opt_lib/"pkgconfig"
     ohai "ImageMagick PKG_CONFIG_PATH: ", imagemagick_lib_path
     ENV.prepend_path "PKG_CONFIG_PATH", imagemagick_lib_path
+    ENV.append "CFLAGS", " -DFD_SETSIZE=10000 -DDARWIN_UNLIMITED_SELECT "
```
d12frosted, line 60: you don't need these extra spaces around

d12frosted (changes requested):

Formula/emacs-plus@26.rb
jixiuf: > you don't need these extra spaces around

fixed.

d12frosted (approved):

PR #510, opened 2022-09-28
add --with-poll

eeshugerman:
Adds a patch/option based on Robert Pluim's [`feature/more-fds`](https://git.savannah.gnu.org/cgit/emacs.git/log/?h=feature/more-fds) branch. This addresses the same issue as https://github.com/d12frosted/homebrew-emacs-plus/issues/487, but by migrating from `select()` to the more modern and performant `poll()` instead of by bumping `FD_SETSIZE`.

Based on the development guidelines doc, I think I'm supposed to add the patch to https://github.com/d12frosted/emacs-plus-basis? But I didn't see any pull requests there, so wasn't sure. Let me know if there's anything I need to do with that.

### DIFF
--- a/Formula/emacs-plus@29.rb
+++ b/Formula/emacs-plus@29.rb
@@ -20,6 +20,7 @@ class EmacsPlusAT29 < EmacsBase
   option "with-no-frame-refocus", "Disables frame re-focus (ie. closing one frame does not refocus another one)"
   option "with-native-comp", "Build with native compilation"
   option "with-compress-install", "Build with compressed install optimization"
+  option "with-poll", "Experimental: use poll() instead of select() to support > 1024 file descriptors`"
 
   #
   # Dependencies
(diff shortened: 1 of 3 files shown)


d12frosted:
Thank you for your first contribution!

> Adds a patch/option based on Robert Pluim's [feature/more-fds](https://git.savannah.gnu.org/cgit/emacs.git/log/?h=feature/more-fds) branch. This addresses the same issue as https://github.com/d12frosted/homebrew-emacs-plus/issues/487, but by migrating from select() to the more modern and performant poll() instead of by bumping FD_SETSIZE.

Looks good. I don't like the idea of having more options, but since this is experimental, let's include it.

> Based on the development guidelines doc, I think I'm supposed to add the patch to https://github.com/d12frosted/emacs-plus-basis? But I didn't see any pull requests there, so wasn't sure. Let me know if there's anything I need to do with that.

Updating development guidelines is on my list. You don't need to add anything to emacs-plus-basis anymore. Thanks for reading it though 😸

Issue #320, opened 2021-03-07
Rounded corners when no title bar


rougier:
When using the no title bar option, I noticed that the frame does not have rounded corners anymore. After trying to understand why it was the case, I think the mask of the window should be something like [NSWindowStyleMaskFullSizeContentView](https://developer.apple.com/documentation/appkit/nswindowstylemask) and the [NSWindowTitleVisibility](https://developer.apple.com/documentation/appkit/nswindowtitlevisibility?language=objc) should be set to hidden.  However, I'm not sure how this should be combined the the current [no-title-bar patch](https://github.com/d12frosted/homebrew-emacs-plus/blob/master/patches/emacs-28/no-titlebar.patch). 

Relevant links:
- https://github.com/lukakerr/NSWindowStyles
- https://stackoverflow.com/questions/35944792/window-title-bar-appears-transparent-issue-not-really-transparent
- https://gitlab.com/gnachman/iterm2/-/issues/7143

d12frosted:
Hey @rougier 

Nice investigation. You are totally right that the reason is `NSWindowStyles` used by this patch.

>  However, I'm not sure how this should be combined the the current no-title-bar patch.

Ineeded. I don't like this patch for not being composable nor configurable. Ideally this patch should not change `NSWindowStyle`, but instead provide a variable controlling `NSWindowStyle` and this variable should be configurable from `init.el`.

rougier:
That would be awesome (and well beyond my knowledge of Apple internals)

winmillwill:
I agree with not combining it with the current no-title-bar behavior out of pure self-interest. I use this patch with a tiling window manager, and it would bug me to see my desktop through the space left by rounded corners. I could see someone else curating a desktop that is *meant* to show through the rounded corners, but that's not me. So long as the behavior of the `--no-title-bar` flag in brew remains unchanged, I'm happy, and I'd be fine with needing to use a different flag, but having the default just change would be slightly annoying to me. I'd still be fine with it, just very low-key annoyed.

rougier:
Maybe there could be an additional flag `--with-round-corners` or `--no-round-corners`.

cxa:
I'm using this patch, without title bar and keep the rounded corners, works very well:

```diff
diff --git a/src/nsterm.m b/src/nsterm.m
index aa29c13eb2..6cec091b8e 100644
--- a/src/nsterm.m
+++ b/src/nsterm.m
[...]

PR #511, opened 2022-09-28
emacs-28: make corners round no-titlebar option

brorbw:
Yabai now uses round borders by default https://github.com/koekeishiya/yabai/issues/1460. This make it look weird with the `--with-no-titlebar` option. This code is stolen from https://github.com/d12frosted/homebrew-emacs-plus/issues/320#issuecomment-965053374

### DIFF
--- a/Formula/emacs-plus@28.rb
+++ b/Formula/emacs-plus@28.rb
@@ -22,6 +22,7 @@ class EmacsPlusAT28 < EmacsBase
   option "with-ctags", "Don't remove the ctags executable that Emacs provides"
   option "with-x11", "Experimental: build with x11 support"
   option "with-no-titlebar", "Experimental: build without titlebar"
+  option "with-no-titlebar-and-round-corners", "Experimental: build without titlebar and round coners"
   option "with-debug", "Build with debug symbols and debugger friendly optimizations"
   option "with-xwidgets", "Experimental: build with xwidgets support"
   option "with-no-frame-refocus", "Disables frame re-focus (ie. closing one frame does not refocus another one)"
@@ -83,6 +84,7 @@ class EmacsPlusAT28 < EmacsBase
   #
 
   local_patch "no-titlebar", sha: "2fa80efc5cda7e96d88a5d145c9313092a6e53d38825c41967c745f08778c41b" if build.with? "no-titlebar"
+  local_patch "no-titlebar-and-round-corners", sha: "ea9dd629ff4ccf36d893e9ab02ced181cd0a3a62fe1973b0fd6d2ed4ab8a3ee7" if build.with? "no-titlebar-and-round-corners"
   local_patch "no-frame-refocus-cocoa", sha: "fb5777dc890aa07349f143ae65c2bcf43edad6febfd564b01a2235c5a15fcabd" if build.with? "no-frame-refocus"
   local_patch "fix-window-role", sha: "1f8423ea7e6e66c9ac6dd8e37b119972daa1264de00172a24a79a710efcb8130"
   local_patch "system-appearance", sha: "d6ee159839b38b6af539d7b9bdff231263e451c1fd42eec0d125318c9db8cd92"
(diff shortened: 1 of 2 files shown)


d12frosted:
Hey @brorbw, thanks for the PR. I will check it later this week. 

Keep in mind that `no-title-bar` option is available only for Emacs 28, and starting with Emacs 29 (that will release next year?) there is no such thing. Built-in solution results in rectangles.

P.S. I also need to upgrade to latest yabai to see it myself 😄

brorbw:
I only applied this change to Emacs 28, because doom is not yet supported on Emacs 29.

>Built-in solution results in rectangles

Right. Maybe I would still be possible to modify the NSWindow to make corners round. Or yabai includes a fix for square corners 🤷🏻

brorbw:
cxa made an update that enables round corners for 29
https://github.com/d12frosted/homebrew-emacs-plus/issues/320#issuecomment-1261831899

d12frosted:
> I only applied this change to Emacs 28, because doom is not yet supported on Emacs 29.

Yeah, that's a good start. I am just thinking about near future where Emacs 29 is released. Not having an option for Emacs 29 would stop some people from upgrading. And it also would create a dept on Emacs+.

I am glad that there is a solution for Emacs 29.

I tested your PR and it looks good. I have only one request. This is a breaking change, so for Emacs 28 I would prefer two options: `--with-no-title-bar` and `--with-no-title-bar-round`. README needs to updated as well.

It would be nice to add `--with-not-tile-bar-round` for Emacs 29, but this can be part of a separate PR.

brorbw:
I have also considered if the option should be conditional. Where you have `--with-no-titlebar` and `--with-round-corners`. I think that serves as an elegant solution to the naming problem but it might introduce too much complexity trying to maintain the different patches. 

I thing if the old `--with-no-titlebar` could be migrated to use the patch cxa made, then an additional option for `--with-round-corners` seems more feasible.

d12frosted:
@brorbw sounds cool from the user perspective, but looks to complicated to implement and maintain. Especially since Emacs 29 is getting closer and closer to release, where we need only one patch.

Ideally these things should be configurable from `init.el` not by compilation options. That way the patch can be applied unconditionally and it's users call to use extra stuff from their Emacs configurations. The same as in Emacs 29 you can configure title bar without recompiling Emacs. Another example of such approach is [System appearance change](https://github.com/d12frosted/homebrew-emacs-plus#system-appearance-change).

If someone has desire and knowledge to implement it, it would be a great contribution.

But meanwhile, having an extra option looks good enough :smile:

brorbw:
I agree completely. I think the purist in me wanted the options to be separate. I'll add it as a separate option. What do you think about the more verbose `--with-no-titlebar-and-round-corners`

d12frosted:
You naming suggestion sounds good to me 😄 👍

Formula/emacs-plus@28.rb
```diff
@@ -21,7 +21,8 @@ class EmacsPlusAT28 < EmacsBase
   # Opt-in
   option "with-ctags", "Don't remove the ctags executable that Emacs provides"
   option "with-x11", "Experimental: build with x11 support"
-  option "with-no-titlebar", "Experimental: build without titlebar"
+  option "with-no-titlebar-and", "Experimental: build without titlebar"
```
d12frosted, line 24: Typo :)

Formula/emacs-plus@28.rb
```diff
@@ -82,7 +83,8 @@ class EmacsPlusAT28 < EmacsBase
   # Patches
   #
 
-  local_patch "no-titlebar", sha: "2fa80efc5cda7e96d88a5d145c9313092a6e53d38825c41967c745f08778c41b" if build.with? "no-titlebar"
+  local_patch "no-titlebar", sha: "69c246224fa7dc4bcb84a8286e8f67eb519e2699319010054a9f71a301a9e3bd" if build.with? "no-titlebar"
```
d12frosted, line 86: I think this change is also made by mistake, this patch wasn't touched, so SHA should remain the same.

Formula/emacs-plus@28.rb
brorbw: 🙃

d12frosted (approved):

d12frosted:
Thank you!

d12frosted:
Mentioned this option in docs (together with some screenshots) in #512 and made it impossible to use both options at the same time in #513.

PR #514, opened 2022-10-05
emacs-28: remove toolbar when using no-title-bar-and-round-corners

d12frosted:


patches/emacs-28/no-titlebar-and-round-corners.patch
```diff
@@ -23,3 +42,13 @@ index 1643ba5..4feca66 100644
    return self;
  }
  
+ 
+ - (void)createToolbar: (struct frame *)f
+ {
++  return;
+   if (FRAME_UNDECORATED (f) || !FRAME_EXTERNAL_TOOL_BAR (f))
```
d12frosted, line 49: In general, `return` here is no not needed. But with the round corners patch, the frame is considered decorated, and so we need to make sure that any call to create toolbar noops.

patches/emacs-28/no-titlebar-and-round-corners.patch
```diff
@@ -10,7 +19,17 @@ index 1643ba5..4feca66 100644
                                 | NSWindowStyleMaskResizable         \
                                 | NSWindowStyleMaskMiniaturizable    \
                                 | NSWindowStyleMaskClosable)
-@@ -8323,6 +8324,12 @@ - (instancetype) initWithEmacsFrame:(struct frame *)f
+@@ -8311,9 +8312,6 @@ - (instancetype) initWithEmacsFrame:(struct frame *)f
+       if ([col alphaComponent] != (EmacsCGFloat) 1.0)
+         [self setOpaque:NO];
+ 
+-      /* toolbar support */
```
d12frosted, line 26: strictly speaking, this is not needed

PR #517, opened 2022-10-09
add docs for showing Emacs in spotlight

whompyjaw:
Personally, I love to launch apps through Spotlight. Symlinking (`ln -s`) the `Emacs.app` to `/Applications` did not make the link/app to show in Spotlight (even after reboot). So, I found an example on SO that worked using Finder's `Create Alias` [here](https://apple.stackexchange.com/a/32495/449886). I figured others would like to do this :)

I looked over the Development Guidelines and I didn't see any specific rules on docs, but please let me know if you prefer any other phrasing or ordering :)


README.org
```diff
@@ -336,6 +339,20 @@ Knows issues:
 | =--with-retro-sink-bw-icon=                   | Unknown                    | [[/icons/preview/retro-sink-bw_128.png]]                   | [[https://www.teuton.org/~ejm/emacsicon/][→ Link]]  |
 | =--with-retro-sink-icon=                      | [[https://www.teuton.org/~ejm/][→ Erik Mugele]]              | [[/icons/preview/retro-sink_128.png]]                      | [[https://www.teuton.org/~ejm/emacsicon/][→ Link]]  |
 | =--with-spacemacs-icon=                       | [[https://github.com/nashamri][→ Nasser Alshammari]]        | [[/icons/preview/spacemacs_128.png]]                       | [[https://github.com/nashamri/spacemacs-logo][→ Link]]  |
+** Spotlight Alias
+*** Apple Silicon
```
d12frosted, line 343: Aliases work the same regardless of the platform. So I would remove this distinction.

d12frosted (commented):
I like the idea, but I would add this to caveats section of the formula itself. Even more, I would provide a Apple Script one-liner that will create alias for you. If you need help with that, let me know.

The problem with this proposed documentation is that you need to know installation location, which differs from platform to platform (Apple Silicon and Intel) and from Emacs version.

d12frosted:
Just to explain a little bit more the idea I proposed.

Today we have the following caveats section:

```
==> Caveats
Emacs.app was installed to:
  /opt/homebrew/opt/emacs-plus@29

To link the application to default Homebrew App location:
  ln -s /opt/homebrew/opt/emacs-plus@29/Emacs.app /Applications                                                                                                                                            
```

It is defined in each formula like this:

https://github.com/d12frosted/homebrew-emacs-plus/blob/85c84fd60418aaed3be3972728380ab890311dd3/Formula/emacs-plus%4029.rb#L246-L258

I propose to either completely remove `ln -s` (cause it doesn't work with Spotlight) or to add an extra note to use the following script for Spotlight:

```bash
osascript -e 'tell application "Finder" to make alias file to posix file "/opt/homebrew/opt/emacs-plus@29/Emacs.app" at POSIX file "/Applications"'
```

The good thing is that caveats can be constructed using some path variables:

```bash
osascript -e 'tell application "Finder" to make alias file to posix file "#{prefix}/Emacs.app" at POSIX file "/Applications"'
```

This section appears after you install Emacs+ and when you run `brew info emacs-plus`.

WDYT?

aguynamedben:
+1 to your solution. I ran into this snag and the osascript worked great for me! I usually use "Emacs for OS X", but tried your distro for native-comp + Emacs 29 and it's working awesome!

The symlink solution also doesn't give me the icon, which was confusing, and opens a terminal before it opens Emacs. The alias is great and I think putting it in `brew info` is fair if it can't be done automatically.

Thank you for working on this fantastic version of Emacs ❤️

LukeDRussell (commented):
Works for me!

saltsucker:
> Just to explain a little bit more the idea I proposed.
> 
> Today we have the following caveats section:
> 
> ```
> ==> Caveats
> Emacs.app was installed to:
>   /opt/homebrew/opt/emacs-plus@29
> 
> To link the application to default Homebrew App location:
>   ln -s /opt/homebrew/opt/emacs-plus@29/Emacs.app /Applications                                                                                                                                            
> ```
> 
> It is defined in each formula like this:
> 
> https://github.com/d12frosted/homebrew-emacs-plus/blob/85c84fd60418aaed3be3972728380ab890311dd3/Formula/emacs-plus%4029.rb#L246-L258
> 
> I propose to either completely remove `ln -s` (cause it doesn't work with Spotlight) or to add an extra note to use the following script for Spotlight:
> 
> ```shell
> osascript -e 'tell application "Finder" to make alias file to posix file "/opt/homebrew/opt/emacs-plus@29/Emacs.app" at POSIX file "/Applications"'
> ```
> 
> The good thing is that caveats can be constructed using some path variables:
> 
> ```shell
> osascript -e 'tell application "Finder" to make alias file to posix file "#{prefix}/Emacs.app" at POSIX file "/Applications"'
> ```
> 
> This section appears after you install Emacs+ and when you run `brew info emacs-plus`.
> 
> WDYT?

This looks great! The script worked for me as well (showed up in spotlight). I made changes per your suggestions, but I don't see any reason to include my doc request if the caveat will cover that? I also don't see why you can't just call the osascript automatically in the brew file? I am not very familiar with brew's install process, but I would think you can call this osascript.

saltsucker:
Sorry that I was MIA for so long. I was working so much that I didn't really dev much in personal time (sadly work doesn't use macOS). But, hopefully this is fixed? I still see that it can't be rebased due to conflicts. I thought I resolved them, so not sure. I don't work too much with merging forks into a master branch.

d12frosted:
Fixed by #642

PR #542, opened 2022-12-27
fix(native-comp): Add libgccjit paths to build environment

jidicula:
When installed with Homebrew, libgccjit 12 writes header files to the Homebrew `include` directory and DLLs to Homebrew's `lib/gcc/current` directory. If these locations are not added to `CPATH` and `LIBRARY_PATH`, respectively, the Emacs build toolchain is unable to locate libgccjit, leading to the build failure described in #485.

This change appends those libgccjit paths to CPATH and LIBRARY_PATH for all Emacs versions that support the native compilation build flag.

Resolves: #485

### DIFF
--- a/.github/workflows/emacs-29.yml
+++ b/.github/workflows/emacs-29.yml
@@ -28,7 +28,7 @@ jobs:
         build_opts:
           - ""
           - "--with-xwidgets"
-          # - "--with-native-comp"
+          - "--with-native-comp"
           - "--with-x11"
 
     env:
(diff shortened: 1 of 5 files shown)


jidicula:
@d12frosted I'm also adding a commit to re-enable the `with-native-comp` flag in CI for 29 and 30 - let's see if it turns green with this fix 🙂

d12frosted (approved):
Nice! Thank you very much. I am pretty sure many people are going to enjoy it 🎉

Issue #527, opened 2022-11-23
Support for Tree-Sitter


danielfleischer:
Emacs 29 master is introducing support for [tree-sitter](https://tree-sitter.github.io/tree-sitter/) parsing system. 

There's a new configuration flag `--with-tree-sitter`. Of course the library itself should be installed before, e.g. `brew install tree-sitter`. Thanks!

d12frosted:
Are there any drawbacks of enabling it by default?

danielfleischer:
I think it will fail if it can't find the tree-sitter library installed on your system. I would make it optional and see what happens in about 2 months when version 29 will be officially out.

d12frosted:
We can define dependencies, that's not a problem. I'll need to check implications before adding this feature as I would like to keep number of option close to zero.

d12frosted:
I've started to look into this feature. I didn't find any user documentation on this feature (my search-fu is bad, I confess). And maybe you know where can I read more about this feature (as a user and not Emacs developer)?

Anyways, it seems like installing `tree-sitter` is enough to make this feature enabled by `configure` script. Can you confirm this?

https://github.com/emacs-mirror/emacs/blob/b3c9840066c90dfa051a9f525461914e92a4621d/configure.ac#L466

I also found this - https://github.com/emacs-tree-sitter/elisp-tree-sitter, but seems like it doesn't require anything except for dynamic modules, which are already enabled in Emacs+.

So need more input on this feature request :)

kommen:
>  didn't find any user documentation on this feature (my search-fu is bad, I confess)

It's not you 😀: https://lists.gnu.org/archive/html/emacs-devel/2022-11/msg01558.html

> We are just starting to work on tree-sitter’s user-facing documentation, 
including the (emacs) manual and NEWS file. I’ll definitely improve the error 
message.


> Anyways, it seems like installing tree-sitter is enough to make this feature enabled by configure script. Can you confirm this?

Just tried it here (macOS Ventura on an M1 Mac) and installing tree-sitter via `brew install tree-sitter` first did make the `brew install emacs-plus@29 -vv` pick it up, as the configure script also reports this:

```
checking for tree-sitter >= 0.20.2... no
checking for tree-sitter >= 0.6.3... yes
...
Does Emacs use -ltree-sitter?                           yes
```

And in the running Emacs `(treesit-available-p)` returns `t`.

ymfsing:
[...]

PR #546, opened 2023-02-06
emacs@{29,30}: depend on tree sitter

d12frosted:
See #527


d12frosted:
Adding this unconditionally, because at this point I see no reason someone wants to disable it (remember that I want to keep amount of options to bare minimum). If someone things I am wrong, let me know (even after I merge this PR, we are not on a wedding).

PR #563, opened 2023-04-12
Install info pages Fixes #561

### LINKED ISSUE #561: Emacs info pages are missing
The Emacs (emacs-plus@29) info files are missing when I go to the root node with `C-h i` (then `d` if necessary).

It turns out this is because the `dir` is not generated after install. After running the following I see the info files I was expecting:

```bash
cd "/opt/homebrew/Cellar/emacs-plus@29/29.0.60/share/info/emacs/"
for file in *.info ; do install-info "$file" dir; done
```

It would be nice to include this step when installing Emacs.

marcus233:


d12frosted (approved):
Thank you!

Issue #556, opened 2023-03-02
Could not install emacs-plus@29 with --with-xwidgets --with-imagemagick --with-native-comp


thermopyle:
Please make sure to follow these steps (and mark the checkboxes):

- [ X] run `brew update` and try to reproduce the issue again
- [ X] run `brew doctor`, fix all issues and try to reproduce your issue again
- [ X] run `brew config` and `brew doctor` and include their output

## What you were trying to do

I was trying to (re-)install emacs-plus@29 on a MacBook Pro M1.

## What happened (include command output)
Build failed with the following configure error;

```
checking for libgccjit.h... yes
configure: error: The installed libgccjit failed to compile and run a test program using
the libgccjit library; see config.log for the details of the failure.
The test program can be found here:
<https://gcc.gnu.org/onlinedocs/jit/intro/tutorial01.html>.
You can try compiling it yourself to investigate the issues.
Please report the issue to your distribution if libgccjit was installed
through that.
You can find the instructions on how to compile and install libgccjit from
source on this site:
<https://gcc.gnu.org/wiki/JIT>.

```

Before installation I uninstalled emacs-plus@29. I uninstalled libgccjit and cleared the caches under /Library/Caches/Homebrew for both emacs-plus and libgccjit, then installed libgccjit again and try to install emacs-plus@29 but with the same error.

<pre>
$ brew install -vv emacs-plus@29 --with-xwidgets --with-imagemagick --with-native-comp
</pre>

<details>
  <summary>Command output</summary>
  <pre>

  ➜  ~ brew install -vv emacs-plus@29 --with-xwidgets --with-imagemagick --with-native-comp
PATH value was changed to:
/opt/homebrew/opt/make/libexec/gnubin
/opt/homebrew/opt/autoconf/libexec/gnubin
/opt/homebrew/opt/gnu-sed/libexec/gnubin
/opt/homebrew/opt/gnu-tar/libexec/gnubin
/opt/homebrew/opt/awk/libexec/gnubin
/opt/homebrew/opt/coreutils/libexec/gnubin
/opt/homebrew/opt/pkg-config/libexec/gnubin
/opt/homebrew/opt/texinfo/libexec/gnubin
/opt/homebrew/opt/xz/libexec/gnubin
/opt/homebrew/opt/gnutls/libexec/gnubin
/opt/homebrew/opt/librsvg/libexec/gnubin
/opt/homebrew/opt/little-cms2/libexec/gnubin
[...]

PR #571, opened 2023-05-03
Fix grep build breakage

jidicula:
# Main change

I found that on my system, I had a different incompatible version of grep from a
plan9 installation on my `PATH` that was superseding the builtin grep on one of my
Macs. This explains why the `with-native-comp` build was [failing on one of my
Macs but not the other](https://github.com/d12frosted/homebrew-emacs-plus/issues/555#issuecomment-1527665327) - the Mac with the successful emacs-plus@28
`with-native-comp` build only had the built-in version of grep on its PATH.

Adding GNU grep as a build-time dependency will include it in the `gnubin`
directory, which is prepended to the PATH in each formula. I've also added an
explicit PATH prepending step for each of the GNU build-time dependencies. This
indicates the coupling of those dependencies with the `gnubin` PATH prepending.

The smoking gun in the `0.6.configure` logs is:

```
usage: grep [-bchiLlnsv] [-e pattern] [-f patternfile] [file ...]
dirname: missing operand
```

I've added extra logging to `Library/EmacsBase.rb` to indicate which version of grep is being used at build-time.

Related to https://github.com/d12frosted/homebrew-emacs-plus/issues/555 and https://github.com/d12frosted/homebrew-emacs-plus/issues/556

# Other related fixes for `gcc` lib linkage consistency
* Adds a fix for appending `gcc`'s libs to `CPATH` and `LIBRARY_PATH` with a colon. 
* Appends `gcc/current` to `LDFLAGS`.

### DIFF
--- a/Formula/emacs-plus@28.rb
+++ b/Formula/emacs-plus@28.rb
@@ -130,6 +130,7 @@ class EmacsPlusAT28 < EmacsBase
     # Necessary for libgccjit library discovery
     ENV.append "CPATH", "#{HOMEBREW_PREFIX}/include", ":" if build.with? "native-comp"
     ENV.append "LIBRARY_PATH", "#{HOMEBREW_PREFIX}/lib/gcc/current", ":" if build.with? "native-comp"
+    ENV.append "LDFLAGS", "#{HOMEBREW_PREFIX}/lib/gcc/current", ":" if build.with? "native-comp"
 
     args <<
       if build.with? "dbus"
--- a/Formula/emacs-plus@28.rb
+++ b/Formula/emacs-plus@28.rb
@@ -36,6 +36,7 @@ class EmacsPlusAT28 < EmacsBase
   depends_on "autoconf" => :build
   depends_on "gnu-sed" => :build
   depends_on "gnu-tar" => :build
+  depends_on "grep" => :build
   depends_on "awk" => :build
   depends_on "coreutils" => :build
   depends_on "pkg-config" => :build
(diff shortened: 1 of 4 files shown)


Formula/emacs-plus@28.rb
```diff
@@ -36,6 +36,7 @@ class EmacsPlusAT28 < EmacsBase
   depends_on "autoconf" => :build
   depends_on "gnu-sed" => :build
   depends_on "gnu-tar" => :build
+  depends_on "grep" => :build
```
jidicula, line 39: @d12frosted I wonder if it may be worth explicitly adding _all_ the autoconf GNU dependencies (perhaps in a future PR?) to cover this scenario better.

d12frosted: I think yes. Ideally I would love to find a way to use homebrew `superenv` (e.g. default environment) which has everything properly configured. But path injection requires custom environment and hence all these issues 😢

Formula/emacs-plus@29.rb
```diff
@@ -156,6 +158,8 @@ def install
     args << "--with-xwidgets" if build.with? "xwidgets"
 
     ENV.prepend_path "PATH", Formula["gnu-sed"].opt_libexec/"gnubin"
+    ENV.prepend_path "PATH", Formula["gnu-tar"].opt_libexec/"gnubin"
+    ENV.prepend_path "PATH", Formula["grep"].opt_libexec/"gnubin"
```
d12frosted, line 162: See my comment from #573 regarding this line.

PR #572, opened 2023-05-03
ci: Save and upload Homebrew logs as workflow artifacts

jidicula:
Hopefully this makes it a bit easier to verify correctness of builds in CI, or to debug CI failures.

Homebrew logs will appear as workflow artifacts that are downloadable from the workflow run summary:
image

I've also included the runner OS in the artifact filenames in case the os build matrix is ever expanded.

See also https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

### DIFF
--- a/.github/workflows/emacs-26.yml
+++ b/.github/workflows/emacs-26.yml
@@ -39,3 +39,19 @@ jobs:
 
       - name: Test installation
         run: $(brew --prefix)/bin/emacs --batch --eval='(print (+ 2 2))'
+
+      - name: Pack up build logs
+        run: |
+          BUILD_OPTS=$(echo "${{ matrix.build_opts }}" | sed 's/ //')
+          echo "build_opts=$BUILD_OPTS" >> "$GITHUB_ENV"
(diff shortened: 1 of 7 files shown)


jidicula:
> Oh wow. Didn't know about this. How do you access these uploaded artifacts?

They'll show up in the workflow summary as downloadable artifacts - I'll drop a screencap in here once I get the run to succeed 😁

jidicula:
@d12frosted see screencap in the PR body. Should be ready for review now! 🙌

d12frosted (approved):
oh my god, this looks awesome

d12frosted:
Thank you. IMO it should be really useful to debug some of the build issues. 🙏

jidicula:
@d12frosted I'm realizing that the log upload step never gets run if the previous build or test steps fail, which means we only get logs for successful runs 🤦‍♂️ ([example](https://github.com/d12frosted/homebrew-emacs-plus/actions/runs/4917986658/jobs/8783743058)) . I'll submit another PR to (hopefully) get around this.

PR #574, opened 2023-05-03
fix: Compiler flag appending

jidicula:
For consistency with LIBRARY_PATH and CPATH appending, we should also append to
LDFLAGS at build-time, to ensure that the same version of gcc and its libs are
used throughout the build.

Possibly related to https://github.com/d12frosted/homebrew-emacs-plus/issues/555 and https://github.com/d12frosted/homebrew-emacs-plus/issues/556

### DIFF
--- a/Formula/emacs-plus@28.rb
+++ b/Formula/emacs-plus@28.rb
@@ -133,9 +133,9 @@ class EmacsPlusAT28 < EmacsBase
     ENV.append "CFLAGS", "-DFD_SETSIZE=10000 -DDARWIN_UNLIMITED_SELECT"
 
     # Necessary for libgccjit library discovery
-    ENV.append "CPATH", "#{HOMEBREW_PREFIX}/include" if build.with? "native-comp"
-    ENV.append "LIBRARY_PATH", "#{HOMEBREW_PREFIX}/lib/gcc/current" if build.with? "native-comp"
-    ENV.append "LDFLAGS", "#{HOMEBREW_PREFIX}/lib/gcc/current" if build.with? "native-comp"
+    ENV.append "CPATH", "-I#{Formula["libgccjit"].opt_include}" if build.with? "native-comp"
+    ENV.append "LIBRARY_PATH", "-L#{Formula["libgccjit"].opt_lib}" if build.with? "native-comp"
+    ENV.append "LDFLAGS", "-L#{Formula["libgccjit"].opt_lib}" if build.with? "native-comp"
 
     args <<
       if build.with? "dbus"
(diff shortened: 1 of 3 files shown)


Formula/emacs-plus@28.rb
```diff
@@ -128,8 +128,9 @@ def install
     ENV.append "CFLAGS", "-DFD_SETSIZE=10000 -DDARWIN_UNLIMITED_SELECT"
 
     # Necessary for libgccjit library discovery
-    ENV.append "CPATH", "#{HOMEBREW_PREFIX}/include" if build.with? "native-comp"
-    ENV.append "LIBRARY_PATH", "#{HOMEBREW_PREFIX}/lib/gcc/current" if build.with? "native-comp"
+    ENV.append "CPATH", "#{HOMEBREW_PREFIX}/include", ":" if build.with? "native-comp"
+    ENV.append "LIBRARY_PATH", "#{HOMEBREW_PREFIX}/lib/gcc/current", ":" if build.with? "native-comp"
+    ENV.append "LDFLAGS", "#{HOMEBREW_PREFIX}/lib/gcc/current", ":" if build.with? "native-comp"
```
d12frosted, line 133: I thought `ENV.append` properly separates values. Aside from checking implementation (lol), you can simply add multiple values and print the value of variable to see how it looks like.

Formula/emacs-plus@29.rb
```diff
@@ -192,7 +193,6 @@ def install
 
       # (prefix/"share/emacs/#{version}").install "lisp"
       prefix.install "nextstep/Emacs.app"
-      (prefix/"Emacs.app/Contents").install "native-lisp" if build.with? "native-comp"
```
jidicula, line 195: Not quite sure what's going on here - I removed it to see if it would resolve the [`Failed changing dylib ID` error](https://github.com/d12frosted/homebrew-emacs-plus/actions/runs/4910058508/jobs/8766873854#step:4:282)r. In any case, that error isn't present in the [28 `--native-comp` build](https://github.com/d12frosted/homebrew-emacs-plus/actions/runs/4910302877/jobs/8767342369).

jidicula: @bkmit seems to be [hitting the same error](https://github.com/d12frosted/homebrew-emacs-plus/actions/runs/4882653342/jobs/8712988197?pr=576#step:4:256) in #576 🤔

jidicula: Seems to be resolved now, considering CI is green ✅

jidicula:
@d12frosted a gentle nudge on this PR too 🙂

d12frosted:
@jidicula 😅 thanks for pining me, my inbox is a bit messy these days

PR #581, opened 2023-05-08
emacs-29: fix round undecorated frame patch

d12frosted:


d12frosted:
oops, merged by mistake 🤦 but should be ok

PR #587, opened 2023-05-17
bump: emacs 29.0.91

cxb811201:


d12frosted:
oh my, I missed thus PR 😢 I am sorry 🙏 
would appreciate an updated PR that uses the latest version - 29.0.92 🙏

PR #590, opened 2023-06-08
feat: add tree-sitter support

cxb811201:


maddovr:
> Hey,
> 
> This option is [enabled by default in Emacs](https://github.com/emacs-mirror/emacs/blob/46b6d175054e8f6bf7cb45e112048c0cf02bfee9/configure.ac#L467). See #527 for details. Or am I missing something?

You need to have tree-sitter installed, otherwise emacs will be built without it. Same could be said for other things like webp etc...

cxb811201:
> Hey,
> 
> This option is [enabled by default in Emacs](https://github.com/emacs-mirror/emacs/blob/46b6d175054e8f6bf7cb45e112048c0cf02bfee9/configure.ac#L467). See #527 for details. Or am I missing something?

I hadn't noticed this problem, thanks for your reply, I will close this PR

PR #593, opened 2023-07-30
bump: emacs 29.1

dustinfarris:
https://lists.gnu.org/archive/html/emacs-devel/2023-07/msg00879.html

### DIFF
--- a/.github/workflows/emacs-29.yml
+++ b/.github/workflows/emacs-29.yml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-11]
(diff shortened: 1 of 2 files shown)


d12frosted (commented):
Thank you! I am going to merge this PR, but the proper release requires actually changing the source from master branch (which should become `--HEAD` option) to the 29.1 tar. See how it's done in `emacs-plus@29.rb`:

https://github.com/d12frosted/homebrew-emacs-plus/blob/b926ff102067d3864ff4cb8060962ec4d46510ef/Formula/emacs-plus%4028.rb#L4-L12

Would appreciate another PR 🙏

PR #598, opened 2023-08-07
feat: add ahead-of-time native compilation flag

dustinfarris:
Users can now choose between --with-native-comp or --with-native-comp-aot which will set the --with-native-comp compile flag to `yes` or `aot` respectively.


d12frosted:
This is on my table. Will address soon.

PR #600, opened 2023-08-15
Adding webp support to emacs 30

cxb811201:
Since emacs 30 already supports webp natively on mac os.

Detailed information is below:
https://github.com/emacs-mirror/emacs/commit/17d3b70fbfcbb392b9a3e64f1ca05168cd16d3e8

### DIFF
--- a/Formula/emacs-plus@30.rb
+++ b/Formula/emacs-plus@30.rb
@@ -41,6 +41,7 @@ class EmacsPlusAT30 < EmacsBase
   depends_on "little-cms2"
   depends_on "jansson"
   depends_on "tree-sitter"
+  depends_on "webp" => :optional
   depends_on "imagemagick" => :optional
   depends_on "dbus" => :optional
   depends_on "mailutils" => :optional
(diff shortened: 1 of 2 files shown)


d12frosted (commented):
are there any cons for enabling it unconditionally?

Formula/emacs-plus@30.rb
```diff
@@ -158,6 +159,7 @@ def install
 
     args << "--with-modules"
     args << "--with-rsvg"
+    args << "--with-webp" if build.with? "webp"
```
d12frosted, line 162: I mean, maybe it's better to avoid introducing a new option and instead just build with webp support unconditionally.

```suggestion
    args << "--with-webp"
```

cxb811201: Thanks for the heads up, you're right, it would be better to have emacs 30 support webp by default.
I will modify it.

cxb811201: I have modified it, please check it, thanks.

d12frosted:
Thank you for your contribution ❤️

PR #639, opened 2023-12-17
docs: Fix CI status badges and links

jidicula:
The README status badges were previously incorrectly showing a failed status, when the builds are in fact succeeding. I noticed that the URL format seems to have changed since the time those badges were first added, so I updated them and checked that the rendered previews shows the correct statuses. 👇 

image


d12frosted (approved):
Haha 😅  thank you!

Issue #618, opened 2023-09-26
Application is not linked out of box


michaelssingh:
OK,  looks like I might have filled that out correct.

One of the immediate issues that I noticed after the installation was that there wasn't an icon for Emacs in the Application Launcher or Spotlight

On my system trying to create a link in /Application resulted in the following error:
`ln -s /opt/homebrew/opt/emacs-plus@29/Emacs.app /Applications`
Gave an error about read-only file-system

This how I managed to get the icon to appear in the application launcher but it doesn't show up in spotlight
`ln -s /opt/homebrew/opt/emacs-plus@29/Emacs.app /System/Volumes/Data/Applications/Emacs.ap`

When launching Emacs from here the Application Launcher,  why don't we have it call `emacsclient -a emacs` to connect to server that is already running by default.

d12frosted:
Hey @michaelssingh 

Regarding the spotlight/finder - it's a known issue. There were several related issues/prs - #41, #517, #488, #324. I acknowledge that the current instructions are problematic and they do not work. AFAIU the Homebrew-recommended way for apps to appear in applications folder/spotlight/whatever is to use cask (see #41). Unfortunately, I don't have enough time and motivation to tackle this.

I would really appreciate someone to either (a) actually figure out the best way to link the app and update instructions or (b) figure out how to provide cask. 

Regarding the `.app` wrapper for `emacsclient`. https://github.com/d12frosted/homebrew-emacs-plus/issues/398 describes some ways to achieve this. Emacs+ could provide some wrapper out of box (although the question about linking strikes again), but it's low on my priority list (unfortunately).

> When launching Emacs from here the Application Launcher, why don't we have it call emacsclient -a emacs to connect to server that is already running by default.

Specifically the idea of `Emacs.app` to launch `emacsclient` - I am pretty sure that's not what everyone would like to have. Not everyone run the server and more importantly - not everyone want to run the server. That's why I don't touch that 😸

michaelssingh:
> Hey @michaelssingh 

> Regarding the `.app` wrapper for `emacsclient`. #398 describes some ways to achieve this. Emacs+ could provide some wrapper out of box (although the question about linking strikes again), but it's low on my priority list (unfortunately).
> 
> > When launching Emacs from here the Application Launcher, why don't we have it call emacsclient -a emacs to connect to server that is already running by default.
> 
[...]

PR #642, opened 2023-12-21
Fix linking how-to in formula caveats

danking:
Thanks for packaging emacs for those of us on OS X! You're doing a great service for the community. Your work directly enables me to work on github.com/hail-is/hail, so you're only twice removed from actual biomedical research in service of human health 😄 .

I think this supersedes https://github.com/d12frosted/homebrew-emacs-plus/pull/324 and complements https://github.com/d12frosted/homebrew-emacs-plus/pull/517.

It arguably resolves https://github.com/d12frosted/homebrew-emacs-plus/issues/618 .

FWIW, I used this with my Emacs 29.1-1 install and it works fine. I'd have saved a few minutes if this message was available in the caveats rather than needing to search through a few issues.

### DIFF
--- a/Formula/emacs-plus@26.rb
+++ b/Formula/emacs-plus@26.rb
@@ -113,7 +113,7 @@ class EmacsPlusAT26 < EmacsBase
         #{prefix}
 
       To link the application to default Homebrew App location:
-        ln -s #{prefix}/Emacs.app /Applications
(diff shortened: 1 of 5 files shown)


d12frosted (approved):
Great! This looks good 🙃

Thank you for your contribution 🙏❤️

PR #649, opened 2024-01-07
dev: switch to macos-12 on ci

d12frosted:


d12frosted:
ok, Emacs 30 doesn't build - but at least everything else builds fine

PR #657, opened 2024-01-19
bump emacs29 to 29.2

bbenchen:


d12frosted (approved):
I must say, it's lovely to see this PR just right after reading the announcemnet ❤️ 💯 Thanks for contributing.

PR #663, opened 2024-01-26
feat: add alpha-background support to emacs 29 and 30

bbenchen:


s-m-t:
Works on my macOS Ventura 13.1 m1 MacBook Air!

The only issue I've found is that it seems the quartz compositor is doing something funny and creates ghosting issues if you set opacity to 0. I think this can be solved by forcing it to be from 1-100 rather than 0-100 in ns_set_alpha_background

d12frosted:
Oh my. I completely forgot about this PR. Apologies @bbenchen 😿 

I want to limit amount of patches we include as it makes maintenance a bit harder. But I want to find a way to easily apply patches like these. And maybe link them from Emacs+ readme so users can find them and use.

logc:
I would be interested in applying this patch on my local installation. Currently there is no information on how to do this on README, right? Is there an easy way to do it, other than creating your own formula based on this branch? 

Thanks @d12frosted for the maintenance work, and to @bbenchen too for proposing these changes, I would love to have alpha-background support on my emacs!

d12frosted:
@logc Nope, we do not provide any extra functionality (yet?) to help with patches. Since it's just a brew formula, you can always use utilities provided by brew itself. For example, `brew edit` to edit the formula locally. Basically, you need to apply changes from this very PR. Is it easy? Yes. Is it sustainable? No. Unfortunately.

logc:
Right, I was expecting to apply the changes in this PR on my own, without putting any more burden on your project. Yes, maintaining from then on my own mini-fork sounds like a hassle but hey ... maybe I will try 😄 thanks for the hints!

Issue #635, opened 2023-12-08
native-comp is enabled by default in emacs 30, Please add option --with-native-compilation=no


liuyinz:
A clear and concise description of what the desired feature is. If the feature
request is adding existing patch, please share the link to it.

d12frosted:
AFAIU, it's not enabled by default. By default it checks if you have libgccjit installed, and if yes, Emacs will be compiled with native compilation feature. Since I've switched back to super env, your custom installed packages should not leak anymore, so no changes are necessary.

Let me know if the issue still persists,

PR #667, opened 2024-02-28
Initial implementation of `--native-compilation` options, enabling AOT or disabling native compilation

stradicat:
I've implemented and tested a rudimentary way of exposing `--native-compilation=aot` to compile Emacs' `.elc` files at build time, as previously asked in https://github.com/d12frosted/homebrew-emacs-plus/issues/556#issuecomment-1565423124, by passing the new `--with-native-comp-aot` argument I've inserted in the formulae.

I've probably repeated myself in Ruby, as I'm not a Ruby dev, and it could surely welcome some deduplication in further commits.

However, AOT compilation works.


Formula/emacs-plus@28.rb
```diff
@@ -67,6 +68,14 @@ class EmacsPlusAT28 < EmacsBase
...
   end
 
+  if build.with? "native-comp-aot"
+    depends_on "libgccjit" => :recommended
+    depends_on "gcc" => :build
+    depends_on "gmp" => :build
+    depends_on "libjpeg" => :build
+    depends_on "zlib" => :build
+  end
+
```
d12frosted, line 78: You can use `||` operator:

```suggestion
  if build.with? "native-comp" || build.with? "native-comp-aot"
    depends_on "libgccjit" => :recommended
    depends_on "gcc" => :build
    depends_on "gmp" => :build
    depends_on "libjpeg" => :build
    depends_on "zlib" => :build
  end
  
```

Formula/emacs-plus@28.rb
```diff
@@ -203,6 +217,7 @@ def install
       # (prefix/"share/emacs/#{version}").install "lisp"
       prefix.install "nextstep/Emacs.app"
       (prefix/"Emacs.app/Contents").install "native-lisp" if build.with? "native-comp"
+      (prefix/"Emacs.app/Contents").install "native-lisp" if build.with? "native-comp-aot"
```
d12frosted, line 220: same here

d12frosted (commented):
Thanks a lot for the PR.

I wonder - does it even make sense to have non-aot native compilation?

Formula/emacs-plus@28.rb
stradicat: I had considered it, but since it was a late-night idea, I wasn't sure 🥲 
I'll refactor these bits.

About having non-aot native compilation, perhaps it won't hurt users with very fast machines (e.g. Apple Silicon and Intel 10th+ gen processors), where background native compilation won't hurt operative performance.

For slower machines, AOT moving native compilation time to the build process gives native-comp of ELPA packages a bit of a break.

There may be as many use cases as Emacs users, to each their own reasons; to all, an extra option to decide according to their needs.

I'll get back to you soon with the proposed deduplications.

d12frosted: I'd pretty much prefer one option, because it's easier to support.

In case we decide to keep two, they should be marked as mutually exclusive using odie. Like this:

https://github.com/d12frosted/homebrew-emacs-plus/blob/d5dd64d3ad95b5e99c282f8cb7c596ed69da108b/Formula/emacs-plus%4030.rb#L70-L74

d12frosted:
Ah, it would be also nice to consider https://github.com/d12frosted/homebrew-emacs-plus/issues/635 while working on AOT.

This sounds like there are 3 options that Emacs can work with: no native compilation, native compilation, native compilation with AOT. 

Homebrew only allows flags as options, meaning that we can't easily encode 3 states :) We can go with one opt out option called `--without-native-comp` and opt-in/opt-out `--with[out]-native-comp-aot`. Enabling second one requires the first one enabled (i.e. use `odie`).

stradicat:
Good point! I understand why some users may still want Emacs' elisp `byte-compile` to `.elc` without native compilation (e.g. myself on my i386 Pentium with 2Gb of RAM), to avoid the overhead.

I'll take the opportunity to include such an option as well, while I'm at it.

devcexx:
Hi,

I just saw this PR and tried locally for building my own Emacs 29 installation. However, just after building it I got this error when I tried to run `emacs --version`:

```
Error using execdir /usr/local/Cellar/emacs-plus@29/29.2/Emacs.app/Contents/MacOS/:
emacs: dlopen(/usr/local/Cellar/emacs-plus@29/29.2/Emacs.app/Contents/MacOS/../native-lisp/29.2-f364439d/preloaded/register-80045398-3789d1c4.eln, 0x0001): tried: '/usr/local/Cellar/emacs-plus@29/29.2/Emacs.app/Contents/MacOS/../native-lisp/29.2-f364439d/preloaded/register-80045398-3789d1c4.eln' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/local/Cellar/emacs-plus@29/29.2/Emacs.app/Contents/MacOS/../native-lisp/29.2-f364439d/preloaded/register-80045398-3789d1c4.eln' (no such file), '/usr/local/Cellar/emacs-plus@29/29.2/Emacs.app/Contents/MacOS/../native-lisp/29.2-f364439d/preloaded/register-80045398-3789d1c4.eln' (no such file)
```

I built it with the following command:

```
git clone https://github.com/stradicat/homebrew-emacs-plus/
cd homebrew-emacs-plus
brew install --build-from-source Formula/emacs-plus@29.rb --with-native-comp-aot --with-modern-black-variant-icon
```

For fixing the error I needed to manually copy the prebuilt eln files into the .app bundle by just doing:

`cp -r /usr/local/Cellar/emacs-plus@29/29.2/lib/emacs/29.2/native-lisp /usr/local/Cellar/emacs-plus@29/29.2/Emacs.app/Contents/`

Maybe some extra changes in the Formula needs to be done in order to copy that folder to the .app bundle? Or did I do something wrong during the build process?

stradicat:
Thanks for bringing it up! I'll test again in the same way you did.

stradicat:
My Mac just died yesterday, so please hold this PR until I can repair the machine and continue testing to ensure that it works 100% despite having passed all checks.

stradicat:
Hm, I had saved a warning I got last time:

```
Error: Failed changing dylib ID of /usr/local/Cellar/emacs-plus@29/29.2/Emacs.app/Contents/native-lisp/29.2-8ccb7056/c-ts-mode-a80983a9-e27a6682.eln
  from c-ts-mode-a80983a9-e27a6682.eln
    to /usr/local/opt/emacs-plus@29/Emacs.app/Contents/native-lisp/29.2-8ccb7056/c-ts-mode-a80983a9-e27a6682.eln

Error: Failed to fix install linkage

The formula built, but you may encounter issues using it or linking other
formulae against it.
```

This looks odd.

stradicat:
Alright, got the machine running.

Regarding
```
Error: Failed changing dylib ID of /usr/local/Cellar/emacs-plus@29/29.2/Emacs.app/Contents/native-lisp/29.2-8ccb7056/c-ts-mode-a80983a9-e27a6682.eln
  from c-ts-mode-a80983a9-e27a6682.eln
    to /usr/local/opt/emacs-plus@29/Emacs.app/Contents/native-lisp/29.2-8ccb7056/c-ts-mode-a80983a9-e27a6682.eln
Error: Failed to fix install linkage
The formula built, but you may encounter issues using it or linking other
formulae against it.
```

I copied `/usr/local/Cellar/emacs-plus@29/29.2/Emacs.app/Contents/native-lisp/29.2-8ccb7056/c-ts-mode-a80983a9-e27a6682.eln` to `/usr/local/opt/emacs-plus@29/Emacs.app/Contents/native-lisp/29.2-8ccb7056/`, but the file was already there (`[files] are identical (not copied)`).

Therefore, the warnings about missing `.eln` files may be a false positive.
However, I agree that they may look confusing.

stradicat:
> Thanks a lot for the PR.
> 
> I wonder - does it even make sense to have non-aot native compilation?

Just came across across this comment by Eli Zaretskii (GNU Emacs maintainer):

[_"Why are you compiling all of your Lisp files ahead of time?_](https://www.reddit.com/r/emacs/comments/tfd5a7/is_it_worth_using_doom_emacs_with_native/) _That is not how native compilation is supposed to be used. You are supposed to let Emacs compile each file in the background if and when it is loaded. This way, you don't need to wait for the compilation, and also, you never compile files you don't use."_

The man knows his beast. Perhaps I should just close this PR and avoid the hassle.

devcexx:
In my particular case, I used this PR for building my emacs installation because I was having huge performance issues while the native compilation was happening asynchronously. For me it was just a way of offloading this extra process that was impacting my emacs usage to the build time.

Anyway at the end of the day the good thing about this PR is that it adds an optional feature anyway, it won't impact users that doesn't use it. I see it as a nice to have.

stradicat:
In that case, I'll do some more testing.

Issue #668, opened 2024-03-07
d12frosted/emacs-plus/emacs-plus@30: private method `new' called for Tap:Class ~ 


LuciusChen:
### Version

emacs-plus@30

### Make sure to follow these steps before submitting the issue

- [X] run `brew update` and try to reproduce the issue again
- [X] run `brew doctor`, fix all issues and try to reproduce your issue again
- [X] try to reproduce your issue without any custom options to the formula

### What happened?

<pre>
$ brew install emacs-plus@30   
</pre>

<details>
  <summary>Command output</summary>
  <pre>

d12frosted/emacs-plus/emacs-plus@30: private method `new' called for Tap:Class

  </pre>
</details>

### brew doctor output

Your system is ready to brew.

### brew config output

HOMEBREW_VERSION: 4.2.11-54-g76c7245
ORIGIN: https://github.com/Homebrew/brew
HEAD: 76c7245d10d4f6219a2c3ac9b584dd8d9661087d
Last commit: 9 hours ago
Core tap JSON: 07 Mar 02:14 UTC
Core cask tap JSON: 07 Mar 02:14 UTC
HOMEBREW_PREFIX: /opt/homebrew
HOMEBREW_CASK_OPTS: []
HOMEBREW_MAKE_JOBS: 14
HOMEBREW_SORBET_RUNTIME: set
Homebrew Ruby: 3.1.4 => /opt/homebrew/Library/Homebrew/vendor/portable-ruby/3.1.4/bin/ruby
CPU: 14-core 64-bit arm_palma
Clang: 15.0.0 build 1500
Git: 2.44.0 => /opt/homebrew/bin/git
Curl: 8.4.0 => /usr/bin/curl
macOS: 14.3.1-arm64
CLT: 15.3.0.0.1.1708646388
Xcode: 15.3
Rosetta 2: false

stschaef:
same issue 
```
HOMEBREW_VERSION: 4.2.11-54-g76c7245
ORIGIN: https://github.com/Homebrew/brew
HEAD: 76c7245d10d4f6219a2c3ac9b584dd8d9661087d
Last commit: 10 hours ago
Core tap HEAD: 9bcd9bbcc66cd9c7484bb787aa7a0d33b778fffb
Core tap last commit: 12 months ago
Core tap JSON: 07 Mar 03:01 UTC
Core cask tap HEAD: 4aa9fca50259ccb5c709be738120fe09281126cf
[...]

PR #669, opened 2024-03-08
Use Tap.fetch instead of Tap.new

soonhokong:
https://github.com/Homebrew/brew/pull/16834 made Tap::new private and this PR address the issue by using Tap.fetch. Close https://github.com/d12frosted/homebrew-emacs-plus/issues/668 .


d12frosted (approved):
LGTM thanks

PR #673, opened 2024-03-11
Update osascript command to set alias name to Emacs.app

gsingh93:
I'm not sure if the original behavior was intentional, but the `osascript` command in the post install message was creating a symlink called `/Applications/Emacs` instead of `/Applications/Emacs.app`, so I had to manually rename it after. I updated the command so it sets the alias name to `Emacs.app`.

(Disclaimer: I don't know AppleScript, I just asked ChatGPT)

### DIFF
--- a/Formula/emacs-plus@26.rb
+++ b/Formula/emacs-plus@26.rb
@@ -116,7 +116,7 @@ class EmacsPlusAT26 < EmacsBase
         #{prefix}
 
       To link the application to default Homebrew App location:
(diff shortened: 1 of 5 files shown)


d1egoaz:
I tested it on my machine ✅ 

```
at 09:09:59 AM ❯ osascript -e 'tell application "Finder" to make alias file to posix file "/opt/homebrew/opt/emacs-plus@30/Emacs.app" at POSIX file "/Applications"  with properties {name:"Emacs.app"}'
alias file Emacs.app of folder Applications of startup disk

```

d12frosted (approved):
LGTM

Thanks for PR

Issue #692, opened 2024-06-23
[Feature request]: Emacs 31.0.50 bumped


vinurs:
### Feature request description

there need `brew install emacs-plus@31`
![CleanShot 2024-06-23 at 22 28 59@2x](https://github.com/d12frosted/homebrew-emacs-plus/assets/37489300/d007a983-c120-4641-a211-eedd75c47229)

aaronjensen:
Related -- how many versions of Emacs should be supported in this repo? When 31 is added, should the oldest one be dropped?

dertuxmalwieder:
I very much second that. I mean, it's great to have older versions under certain circumstances, but I don't think that @d12frosted should have to maintain all of them forever.

d12frosted:
> Related -- how many versions of Emacs should be supported in this repo? When 31 is added, should the oldest one be dropped?

A good question. I don't mind formulas staying here, but I pretty much want to kill CI workflows. I think it makes sense to care only bout current release, next release candidate if available and master. So normally we are talking about two versions and during transition/release of new version (like now) there are three versions.

d12frosted:
> I mean, it's great to have older versions under certain circumstances, but I don't think that @d12frosted should have to maintain all of them forever.

❤️ I appreciate what you say. Frankly, the most 'complicated' features from maintainer point of view are not related to Emacs version per se, but rather to (a) native compilation and (b) environment of the builds. 

https://github.com/d12frosted/homebrew-emacs-plus/pull/695

WDYT?

aaronjensen:
I think that my general concern with that approach is that it seems to send a mixed message. If it's not under CI, is it really maintained? If anything is wrong (like after a new version of macOS is released), would you or someone fix it? How many people are relying on the older versions? Could they get them from somewhere else like https://emacsformacosx.com/builds instead?

Personally, I'd be inclined to only maintain the most recent version and development versions and include a link to https://emacsformacosx.com/builds in the README. Maintaining OSS is enough work 😄

d12frosted:
@aaronjensen good point. I mean, as long as it works - I don't care. As long as something needs to be done - I am out :D And yes, we can point to emacsformacosx, though it's a different build anyways as Emacs+ still adds a bit of stuff 😅 

So let's wait for public release of the new version and decide what to do.

PR #695, opened 2024-06-28
remove old versions from CI

d12frosted:
See https://github.com/d12frosted/homebrew-emacs-plus/issues/692#issuecomment-2195696280


dertuxmalwieder:
LGTM. :)

PR #699, opened 2024-07-07
fix(emacs-30&native-comp): remove unnecessary symbolic links of native-lisp

ilyaw39:
This PR most likely resolves a potential issue in #686. According to [CODE](https://github.com/d12frosted/homebrew-emacs-plus/blob/af4df5166b357668f8207fbdff46a802633f1a5d/Formula/emacs-plus%4030.rb#L220), the `native-lisp` files have already been created, but #686 performed an unnecessary operation —— As seen in the screenshot, an unnecessary symbolic link was created.

screenshot

### DIFF
--- a/Formula/emacs-plus@30.rb
+++ b/Formula/emacs-plus@30.rb
@@ -281,9 +281,6 @@ class EmacsPlusAT30 < EmacsBase
     Dir.glob(emacs_info_dir/"*.info") do |info_filename|
       system "install-info", "--info-dir=#{emacs_info_dir}", info_filename
     end
-    if build.with? "native-comp"
-      ln_sf "#{Dir[opt_prefix/"lib/emacs/*"].first}/native-lisp", "#{opt_prefix}/Emacs.app/Contents/native-lisp"
-    end
   end
 
   def caveats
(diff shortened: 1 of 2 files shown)


ilyaw39:
> This PR most likely resolves a potential issue in #686. According to [CODE](https://github.com/d12frosted/homebrew-emacs-plus/blob/af4df5166b357668f8207fbdff46a802633f1a5d/Formula/emacs-plus%4030.rb#L220), the `native-lisp` files have already been created, but #686 performed an unnecessary operation —— As seen in the screenshot, an unnecessary symbolic link was created.
> 
> screenshot

I just noticed that this issue has already been addressed at bbenchen#1. We found the same problem, but our solutions are different. bbenchen#1 kept the `ln_sf` part and avoided reinstalling `native-lisp` in the `Emacs.app` directory. Compared to my solution, this reduces file duplication and improves installation speed.

However, I considered that the `ln_sf` part of the code skips the check `if (build.with? "cocoa") && (build.without? "x11")`, which means that even if `Emacs.app` is not correctly generated (for example, if the user selects both `--with-x11` and `--with-native-comp`), the `ln_sf` part will still be executed, leading to installation errors.

Did you have a better solution for this? @bbenchen

bbenchen:
Hi @ilyaw39
I haven't tested x11 and native-comp scenarios, so I don't have any solutions for now.

ilyaw39 (commented):
This solution should be good enough.

d12frosted (approved):
Oh yeah, I was looking at this part recently and could not understand why it was added here since we already cover that in the install flow.

d12frosted:
Thanks @ilyaw39 ❤️

Issue #702, opened 2024-08-01
[Build Failure]: --with-poll fails to apply poll.patch


hazardfn:
### Version

emacs-plus@30

### Make sure to follow these steps before submitting the issue

- [X] run `brew update` and try to reproduce the issue again
- [X] run `brew doctor`, fix all issues and try to reproduce your issue again
- [X] try to reproduce your issue without any custom options to the formula

### What happened?

<pre>
$ brew install emacs-plus --with-ctags --with-dbus --with-mailutils --with-xwidgets --with-imagemagick --with-native-comp --with-poll
</pre>

<details>
  <summary>Command output</summary>
  <pre>

==> Downloading https://formulae.brew.sh/api/formula.jws.json
###################################################################################################################################################################################################################################################### 100.0%
==> Downloading https://formulae.brew.sh/api/cask.jws.json
###################################################################################################################################################################################################################################################### 100.0%
==> Fetching d12frosted/emacs-plus/emacs-plus@30
==> Cloning https://github.com/emacs-mirror/emacs.git
Updating /Users/howard/Library/Caches/Homebrew/emacs-plus@30--git
From https://github.com/emacs-mirror/emacs
   ceb5a152227..cb421286d2e  emacs-30   -> origin/emacs-30
==> Checking out branch emacs-30
Already on 'emacs-30'
Your branch is behind 'origin/emacs-30' by 2 commits, and can be fast-forwarded.
  (use "git pull" to update your local branch)
HEAD is now at cb421286d2e Fix c++-ts-mode indentation for templace (bug#72263)
==> Installing emacs-plus@30 from d12frosted/emacs-plus
==> Patching
==> Applying fix-window-role.patch
patching file 'src/nsterm.m'
==> Applying system-appearance.patch
patching file 'src/frame.h'
patching file 'src/nsfns.m'
patching file 'src/nsterm.m'
==> Applying poll.patch
patching file configure.ac
patching file 'src/nsterm.m'
patching file 'src/process.c'
1 out of 61 hunks failed--saving rejects to 'src/process.c.rej'
patching file 'src/sysdep.c'
patching file 'src/syspoll.h'
patching file 'src/sysselect.h'

  </pre>
</details>

`poll.patch` no longer seems to work with the latest changes on emacs@30.

### brew doctor output

Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
[...]

PR #703, opened 2024-08-03
Fix poll patch

daviderestivo:
Fixes https://github.com/d12frosted/homebrew-emacs-plus/issues/702

### DIFF
--- a/Formula/emacs-plus@30.rb
+++ b/Formula/emacs-plus@30.rb
@@ -99,7 +99,7 @@ class EmacsPlusAT30 < EmacsBase
   opoo "The option --with-no-frame-refocus is not required anymore in emacs-plus@30." if build.with? "no-frame-refocus"
   local_patch "fix-window-role", sha: "1f8423ea7e6e66c9ac6dd8e37b119972daa1264de00172a24a79a710efcb8130"
   local_patch "system-appearance", sha: "9eb3ce80640025bff96ebaeb5893430116368d6349f4eb0cb4ef8b3d58477db6"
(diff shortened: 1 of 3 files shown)


sauvala:
Thanks @daviderestivo! I was also looking into this and created my own patch file https://github.com/d12frosted/homebrew-emacs-plus/pull/704. It is missing the other changes you updated because I only tried to get the patch to process.c file to work. I noticed that our changes are slightly different, mainly because of the first line in each of the hunks in the changes to process.c file, for example:

https://github.com/d12frosted/homebrew-emacs-plus/blob/65c706b7bb68c36bba31bff21ef297c20c6bf184/patches/emacs-30/poll.patch#L87

vs.

https://github.com/d12frosted/homebrew-emacs-plus/blob/7b27b590ebacfd37a1cda1458f75f30ecf9d281a/patches/emacs-30/poll.patch#L86

Do you know why our patches are different? It is very likely that I messed something up with my patch.

daviderestivo:
I have just update the patch since the syspoll.h was missing.

daviderestivo:
> Do you know why our patches are different? It is very likely that I messed something up with my patch.

Dunno. I have just applied the patch to the emacs master branch and fixed the only rejected hunk.

sauvala:
> > Do you know why our patches are different? It is very likely that I messed something up with my patch.
> 
> I have just applied the patch to the emacs master branch and fixed the only rejected hunk. I have generated the patch using "git diff" this might be the reason.

I was applying my batch to the emacs-30 release branch so that would indeed explain the differences. I wonder if we should use emacs-30 release branch to make the patch that targets version 30.

daviderestivo:
I tested it both with emacs-30 and emacs-31 (master) and it seems to work fine for both.

d12frosted (approved):

Issue #706, opened 2024-08-08
[Feature request]:  add --with-mps build option


nailuoGG:
### Feature request description

If we make it easier for more people to use the `scratch/cgi` [branch](https://github.com/emacs-mirror/emacs/tree/scratch/igc) with MPS support, they may be able to test it and provide feedback, which could help that branch develop further.

I've tried modifying the `emacs-plus` formula code to work with `scratch/cgi` branch , and I was able to successfully build a product that can handle daily work, but it still doesn't support some build options.

The options that cannot be compiled (now) are: `--with-xwidgets` and `--with-poll`

The command that can be compiled is:

```bash
brew install emacs-plus@31 --with-imagemagick --with-native-comp --with-modern-alecive-flatwoken-icon
```

Note: This build still occasionally crashes.

d12frosted:
Is there any TLDR of what that branch is about? 

> The options that cannot be compiled (now) are: --with-xwidgets and --with-poll

I think at some point XWidgets Support will be dropped. It's experimental and AFAIK no that many people use it.

> Note: This build still occasionally crashes.

I wonder why.

nailuoGG:
> Is there any TLDR of what that branch is about?

TL;DR: Replace Emacs' built-in GC with a [better one](https://github.com/Ravenbrook/mps). 

Progress: 
- [STatus of MPS branch](https://mail.gnu.org/archive/html/emacs-devel/2024-04/msg00434.html)
- [MPS: master merged](https://lists.gnu.org/archive/html/emacs-devel/2024-08/msg00111.html) Note: merged `master` into `scratch/igc`




> The options that cannot be compiled (now) are: --with-xwidgets and --with-poll

 The build process fails when combining -`-with-xwidgets` or `--with-poll` with `--with-mps`.  
 
>  Note: This build still occasionally crashes.

Error logs were not captured or saved. 

I was using the igc branch built with --with-mps approximately three days before opening this issue.

d12frosted:
We had some discussion in #707. As result, let's see how development of this feature evolves and we can reconsider adding this option later. But for now, I am closing this as there are no plans to add this option.

PR #707, opened 2024-08-09
add --with-mps option

nailuoGG:
build with MPS garbage collector, using scratch/igc branch as default   #706 

```
brew install emacs-plus@31 \
    --with-mps \
    --with-imagemagick  \
    --with-native-comp \
    --with-modern-alecive-flatwoken-icon
```

tip:

set `--with-mps `  will use `scratch/cgi` as default branch.

set `HOMEBREW_EMACS_PLUS_31_BRANCH=scratch/igc` if merged to `other` branch

### DIFF
--- a/.github/workflows/emacs-31.yml
+++ b/.github/workflows/emacs-31.yml
@@ -30,6 +30,7 @@ jobs:
           - "--with-xwidgets"
           - "--with-native-comp"
           - "--with-x11"
+          - "--with-mps"
 
     env:
       HOMEBREW_EMACS_PLUS_MODE: local
--- a/.github/workflows/emacs-31.yml
+++ b/.github/workflows/emacs-31.yml
@@ -31,7 +31,6 @@ jobs:
           - "--with-xwidgets"
           - "--with-native-comp"
           - "--with-x11"
-          - "--with-mps"
 
     env:
       HOMEBREW_EMACS_PLUS_MODE: local
--- a/.github/workflows/emacs-31.yml
+++ b/.github/workflows/emacs-31.yml
@@ -26,24 +26,35 @@ jobs:
       matrix:
         os: [macos-12]
         build_opts:
+          - "--with-mps"
           - ""
           - "--with-xwidgets"
           - "--with-native-comp"
           - "--with-x11"
-          - "--with-mps"
 
     env:
       HOMEBREW_EMACS_PLUS_MODE: local
 
     steps:
-      - uses: actions/checkout@v4
+      - name: Set up Homebrew
+        id: set-up-homebrew
+        uses: Homebrew/actions/setup-homebrew@master
+        with:
+          core: false
+          cask: true
+          test-bot: false
 
+      - name: Enable debug mode
+        run: |
+          echo 'HOMEBREW_DEBUG=1' >> "${GITHUB_ENV}"
+          echo 'HOMEBREW_VERBOSE=1' >> "${GITHUB_ENV}"
+        if: runner.debug
       - name: Install xquartz
         if: contains(matrix.build_opts, '--with-x11')
         run: brew install --cask xquartz
 
       - name: Build emacs-plus@31 ${{ matrix.build_opts }}
-        run: brew install ./Formula/emacs-plus@31.rb ${{ matrix.build_opts }} --verbose
+        run: brew install emacs-plus@31 ${{ matrix.build_opts }} --verbose
 
       - name: Test installation
         run: $(brew --prefix)/bin/emacs --batch --eval='(print (+ 2 2))'
--- a/Formula/emacs-plus@31.rb
+++ b/Formula/emacs-plus@31.rb
@@ -27,6 +27,7 @@ class EmacsPlusAT31 < EmacsBase
   option "with-native-comp", "Build with native compilation"
   option "with-compress-install", "Build with compressed install optimization"
   option "with-poll", "Experimental: use poll() instead of select() to support > 1024 file descriptors`"
+  option "with-mps", "Build With MPS garbage collector."
 
   #
   # Dependencies
@@ -66,6 +67,12 @@ class EmacsPlusAT31 < EmacsBase
     depends_on "zlib" => :build
   end
 
+  if build.with? "mps"
+    depends_on "libmps" => :recommended
+    # Set the HOMEBREW_EMACS_PLUS_31_BRANCH environment variable
+    ENV["HOMEBREW_EMACS_PLUS_31_BRANCH"] = ENV.fetch("HOMEBREW_EMACS_PLUS_31_BRANCH", "scratch/igc")
+  end
+
   #
   # Incompatible options
   #
@@ -80,10 +87,10 @@ class EmacsPlusAT31 < EmacsBase
   # URL
   #
 
-  if ENV['HOMEBREW_EMACS_PLUS_31_REVISION']
-    url "https://github.com/emacs-mirror/emacs.git", :revision => ENV['HOMEBREW_EMACS_PLUS_31_REVISION']
+  if ENV["HOMEBREW_EMACS_PLUS_31_REVISION"]
+    url "https://github.com/emacs-mirror/emacs.git", :revision => ENV["HOMEBREW_EMACS_PLUS_31_REVISION"]
   else
-    url "https://github.com/emacs-mirror/emacs.git", :branch => "master"
+    url "https://github.com/emacs-mirror/emacs.git", :branch => ENV.fetch("HOMEBREW_EMACS_PLUS_31_BRANCH", nil) or "master"
   end
 
   #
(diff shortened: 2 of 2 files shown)


.github/workflows/emacs-31.yml
```diff
@@ -26,6 +26,7 @@ jobs:
       matrix:
         os: [macos-12]
         build_opts:
+          - "--with-mps"
```
d12frosted, line 29: this should go after the "no args" build :) 

but in reality, I am not sure that I care about this build. I mean, if this build fails, I am going to receive emails that I will simply ignore. Plus it loads CI even more. I try to keep the size of build matrix minimal.

.github/workflows/emacs-31.yml
```diff
@@ -35,14 +36,25 @@ jobs:
...
+          echo 'HOMEBREW_DEBUG=1' >> "${GITHUB_ENV}"
+          echo 'HOMEBREW_VERBOSE=1' >> "${GITHUB_ENV}"
+        if: runner.debug
       - name: Install xquartz
         if: contains(matrix.build_opts, '--with-x11')
         run: brew install --cask xquartz
 
       - name: Build emacs-plus@31 ${{ matrix.build_opts }}
-        run: brew install ./Formula/emacs-plus@31.rb ${{ matrix.build_opts }} --verbose
+        run: brew install emacs-plus@31 ${{ matrix.build_opts }} --verbose
```
d12frosted, line 57: Why?

Formula/emacs-plus@31.rb
```diff
@@ -27,6 +27,7 @@ class EmacsPlusAT31 < EmacsBase
   option "with-native-comp", "Build with native compilation"
   option "with-compress-install", "Build with compressed install optimization"
   option "with-poll", "Experimental: use poll() instead of select() to support > 1024 file descriptors`"
+  option "with-mps", "Build With MPS garbage collector."
```
d12frosted, line 30: it must be marked as experimental

```suggestion
  option "with-mps", "Experimental: build With MPS garbage collector."
```

Formula/emacs-plus@31.rb
```diff
@@ -66,6 +67,12 @@ class EmacsPlusAT31 < EmacsBase
     depends_on "zlib" => :build
   end
 
+  if build.with? "mps"
+    depends_on "libmps" => :recommended
+    # Set the HOMEBREW_EMACS_PLUS_31_BRANCH environment variable
+    ENV["HOMEBREW_EMACS_PLUS_31_BRANCH"] = ENV.fetch("HOMEBREW_EMACS_PLUS_31_BRANCH", "scratch/igc")
```
d12frosted, line 73: No need to modify ENV for this. You can simply use different branches based on build options below.

Formula/emacs-plus@31.rb
```diff
@@ -80,10 +87,10 @@ class EmacsPlusAT31 < EmacsBase
   # URL
   #
 
-  if ENV['HOMEBREW_EMACS_PLUS_31_REVISION']
-    url "https://github.com/emacs-mirror/emacs.git", :revision => ENV['HOMEBREW_EMACS_PLUS_31_REVISION']
+  if ENV["HOMEBREW_EMACS_PLUS_31_REVISION"]
+    url "https://github.com/emacs-mirror/emacs.git", :revision => ENV["HOMEBREW_EMACS_PLUS_31_REVISION"]
   else
-    url "https://github.com/emacs-mirror/emacs.git", :branch => "master"
+    url "https://github.com/emacs-mirror/emacs.git", :branch => ENV.fetch("HOMEBREW_EMACS_PLUS_31_BRANCH", nil) or "master"
```
d12frosted, line 93: You can decide which branch to use based on build options.

d12frosted (changes requested):
Thanks for your PR.

There are several minimal changes I'd love to request.

But most importantly, I'd love to understand what this feature is about and why should we increase complexity of the formula for this feature. Over last few years I am trying to remove number of build options :)

I am not familiar with development of this feature, so I have to ask two things. What is the state of its development? And is there a consensus (link needed) that it's going to be merged to master? Or is it simply experimental thing?

nailuoGG:
Thanks for your review. 

> But most importantly, I'd love to understand what this feature is about and why should we increase complexity of the formula for this feature. Over last few years I am trying to remove number of build options :)

**Feature**: Improve Emacs' garbage collection (GC) performance by replacing its built-in GC with a better one from [Ravenbrook/mps](https://github.com/Ravenbrook/mps).

**Why**: Experiment with improved GC performance, more testing needed to make it production-ready. Also simplifies macOS setup and makes it easy to test MPS.

>  What is the state of its development? And is there a consensus (link needed) that it's going to be merged to master? Or is it simply experimental thing?

**Status**: This feature is still experimental, no plans to merge into the master branch soon."

- [STatus of MPS branch](https://mail.gnu.org/archive/html/emacs-devel/2024-04/msg00434.html)
- [MPS: master merged](https://lists.gnu.org/archive/html/emacs-devel/2024-08/msg00111.html) Note: merged `master` into `scratch/igc`

.github/workflows/emacs-31.yml
nailuoGG: > this should go after the "no args" build :)

I need quickly see if everything built successfully with '--with-mps'. so I'll move it up in the build sequence, but afterwards 'no args' should come first. :)

.github/workflows/emacs-31.yml
nailuoGG: Running` brew install ./Formula/emacs-plus@31.rb --with-mps`  thrown error :  

`Error: invalid option: --with-mps`.   [Link](https://github.com/d12frosted/homebrew-emacs-plus/actions/runs/10323706429/job/28581651325#step:4:147)

While same content  place in `/opt/homebrew/Library/Taps/d12frosted/homebrew-emacs-plus/Formula`

Run `brew install emacs-plus@31 --with-mps` work fine

d12frosted: That's strange. I wonder how other options work. I don't remember any details 😂  Need to investigate.

----

*update:*

So I tried adding a new argument in [a separate PR](https://github.com/d12frosted/homebrew-emacs-plus/pull/708) and it seems that CI [picked it up](https://github.com/d12frosted/homebrew-emacs-plus/actions/runs/10465548134/job/28980967099?pr=708#step:4:746) without this change. 

I don't remember details, but it's not really clear to me what exactly is being built when `emacs-plus@31` is referenced instead of specific file.

I also don't really understand why do we need [Setup Homebrew Github Action](https://github.com/Homebrew/actions/tree/master/setup-homebrew). Standard GitHub macos runners have brew installed and configured out of the box. 

> I pulled the workflow from Homebrew-core and it works great.

I think they need it because they are running on ubuntu image which doesn't have homebrew ootb.

d12frosted:
> Improve Emacs' garbage collection (GC) performance by replacing its built-in GC with a better one from [Ravenbrook/mps](https://github.com/Ravenbrook/mps).

Thanks for explanation.

> Why: Experiment with improved GC performance, more testing needed to make it production-ready. Also simplifies macOS setup and makes it easy to test MPS.

That makes sense.

> Status: This feature is still experimental, no plans to merge into the master branch soon.

I see.

Frankly, I am not convinced that adding an extra option to emacs-plus would help developers somehow. Building from separate branch is as trivial as `brew edit emacs-plus@31`. We can simply share a gist with edits necessary. But considering the experimental nature of the feature it might just increase maintenance cost. I can not provide any kind of support for this feature anyways. And having it as an option in emacs-plus (and especially part of build matrix on CI) is a signal that we do support it or at least try to, which is not true.

So I propose to prepare a patch and share it as a gist in #706. If we get enough traction and/or the feature becomes more stable/closer to release - we can add an extra option.

cc @aaronjensen as you have some healthy thoughts on maintenance :)

.github/workflows/emacs-31.yml
nailuoGG: That's weird... if the simplest case works fine, maybe the problem is elsewhere. 

> I also don't really understand why do we need Setup Homebrew Github Action. Standard GitHub macos runners have brew installed and configured out of the box.

I wan to make sure we're testing this formula like a normal user would use brew. but it seems to be bringing in more complexity than necessary.

nailuoGG:
> So I propose to prepare a patch and share it as a gist in https://github.com/d12frosted/homebrew-emacs-plus/issues/706. If we get enough traction and/or the feature becomes more stable/closer to release - we can add an extra option.

I agree that keeping things simple is a good principle, and creating a separate gist for others to use makes sense. 😄

For users, the performance boost from replacing GC might not be immediately noticeable, especially when compared to using something like gcmh.

d12frosted:
At least I am happy that we have a reference for those willing to build with MPS. Let's see how development of this feature evolves and we can reconsider adding this option later.

PR #712, opened 2024-08-25
feat: add missing deps and include super shims to path

d12frosted:


d12frosted:
will reopen in different configuration

PR #717, opened 2024-08-25
feat: enable AOT for native compilation

d12frosted:
fixes #598 but without any extra changes to env


d12frosted:
Tested locally and boy, the experience is much better.

PR #719, opened 2024-08-26
Enable invoking processes using speech recognition APIs

xenodium:
Hi Boris!

Sending a tiny patch to enable invoking processes using speech recognition APIs.

Without NSSpeechRecognitionUsageDescription in Info.plist, we cannot launch processes that use macOS's speech recognition APIs.

For example, running an command line utility from eshell that uses speech recognition yields:

```
$ /tmp/macosrec --ocr --input /tmp/audio.mp3
  abort trap: 6
```

With NSSpeechRecognitionUsageDescription in Info.plist, macOS displays the dialog asking the user to grant permission and allows the speech APIs to be used from the process.

Screenshot 2024-08-26 at 19 48 59

I'll see if I can upstream this as well.


d12frosted:
Hi Álvaro 🙃 I am so happy to see you 🔥 

It would be nice to get it upstream, but if not, maybe we should stay consistent and put all permission-related stuff [here](https://github.com/d12frosted/homebrew-emacs-plus/blob/1cd7ba066193bc7e75e1a4771779afb6662b2354/Library/EmacsBase.rb#L81-L85).

d12frosted (approved):
I think we can take it. Let me know if it lands in upstream 🙏

xenodium:
> Hi Álvaro 🙃 I am so happy to see you 🔥

Likewise! Really nice to see ya active here.

> It would be nice to get it upstream

I'll produce an upstream patch shortly. I'm waiting to hear back about my copyright assignment (unsure how long it'll take).

> maybe we should stay consistent and put all permission-related stuff here.

Yeah. That's better.

>  I think we can take it.

Thank you! You merged before I even replied ;) Was gonna do the upstream patch first.

>  Let me know if it lands in upstream 🙏

Will do!

xenodium:
> Let me know if it lands in upstream 🙏

Submitted patch https://debbugs.gnu.org/cgi/bugreport.cgi?bug=72827

xenodium:
Heads-up, this is now in master https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=c36ff7d8e46c75cfadea6f4c1f4eadb3aef37f2d

Any way we can keep the PR in emacs-plus for Emacs 29 and 30 only? Do we use patch instead? Something else?

d12frosted:
Congratulations on your contribution to GNU Emacs 🎉 

Re your question - already done in https://github.com/d12frosted/homebrew-emacs-plus/commit/6f099bb618334885f509c907efc9a0e0e4257ac5. Not ideal way, but at least it doesn't break 31 for now.

xenodium:
> Congratulations on your contribution to GNU Emacs 🎉

Thanks =) My first.

> Re your question - already done in https://github.com/d12frosted/homebrew-emacs-plus/commit/6f099bb618334885f509c907efc9a0e0e4257ac5. Not ideal way, but at least it doesn't break 31 for now.

Wonderful. Thank you!

PR #740, opened 2024-10-26
Bump emacs-plus@30 to 30.0.92

zonuexe:
[Emacs 30.0.92 pretest is available](https://lists.gnu.org/archive/html/emacs-devel/2024-10/msg00627.html)

### DIFF
--- a/Formula/emacs-plus@30.rb
+++ b/Formula/emacs-plus@30.rb
@@ -10,7 +10,11 @@ class EmacsPlusAT30 < EmacsBase
   homepage "https://www.gnu.org/software/emacs/"
 
   head do
-    url "https://github.com/emacs-mirror/emacs.git", :branch => "emacs-30"
+    if ENV['HOMEBREW_EMACS_PLUS_30_REVISION']
+      url "https://github.com/emacs-mirror/emacs.git", :revision => ENV['HOMEBREW_EMACS_PLUS_30_REVISION']
+    else
+      url "https://github.com/emacs-mirror/emacs.git", :branch => "emacs-30"
+    end
   end
 
   #
--- a/Formula/emacs-plus@30.rb
+++ b/Formula/emacs-plus@30.rb
@@ -2,7 +2,7 @@ require_relative "../Library/EmacsBase"
 
 class EmacsPlusAT30 < EmacsBase
   init 30
-  version "30.0.91"
+  version "30.0.92"
 
   desc "GNU Emacs text editor"
   homepage "https://www.gnu.org/software/emacs/"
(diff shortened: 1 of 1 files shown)


d12frosted (commented):
oh there are pretests already? happy to hear!

bumping the version itself in this file has no effect on _what_ since we build from `emacs-30` branch anyways (at best, it affects update mechanism, which is also important)

but since there are tarballs available, I'd rather start preparing for proper release:

1. build from tarballs by default (truly versioned build); for example, this is how it's done for Emacs 29:

https://github.com/d12frosted/homebrew-emacs-plus/blob/cc3280dfdf7562dd69c97d0dfa872b1c3313a18a/Formula/emacs-plus%4029.rb#L5-L7

2. add a `--head` option to build from `emacs-30` branch; this is how we do it for Emacs 29:

https://github.com/d12frosted/homebrew-emacs-plus/blob/cc3280dfdf7562dd69c97d0dfa872b1c3313a18a/Formula/emacs-plus%4029.rb#L12-L14

but we also need to keep an ability to override revision like this:

https://github.com/d12frosted/homebrew-emacs-plus/blob/cc3280dfdf7562dd69c97d0dfa872b1c3313a18a/Formula/emacs-plus%4030.rb#L80-L84

---

Can I ask you to perform these changes?

zonuexe:
@d12frosted Thank you for suggestion! I added.

Formula/emacs-plus@30.rb
```diff
@@ -79,8 +85,6 @@ class EmacsPlusAT30 < EmacsBase
 
   if ENV['HOMEBREW_EMACS_PLUS_30_REVISION']
```
d12frosted, line 86: The whole "if" clause should go under `head` :)

Formula/emacs-plus@30.rb
```diff
@@ -2,11 +2,17 @@
...
+  url "https://alpha.gnu.org/gnu/emacs/pretest/emacs-30.0.92.tar.xz"
+  mirror "https://ftpmirror.gnu.org/emacs/emacs-30.0.92.tar.xz"
+  sha256 "d89287bd5a8381bb60e14aab95202377261d43a60d15dc0a61d0d662bc5626be"
 
   desc "GNU Emacs text editor"
   homepage "https://www.gnu.org/software/emacs/"
 
+  head do
+    url "https://github.com/emacs-mirror/emacs.git", :branch => "emacs-30"
+  end
```
d12frosted, line 14:
```suggestion
  head do
    if ENV['HOMEBREW_EMACS_PLUS_30_REVISION']
      url "https://github.com/emacs-mirror/emacs.git", :revision => ENV['HOMEBREW_EMACS_PLUS_30_REVISION']
    else
      url "https://github.com/emacs-mirror/emacs.git", :branch => "emacs-30"
    end
  end
```

d12frosted:
Thank you 🔥

zonuexe:
@d12frosted Thank you for supporting me, I enjoy emacs-plus every day!

PR #743, opened 2024-10-30
AppleScript usage of posix is made consistent

aaronjensen:
I could not find a reason for the inconsistency in these scripts. This stopped some of my teammates in their tracks -- they refused to run it because of the inconsistency, questioning whether or not it was necessary or if there was a problem.


d12frosted (approved):

d12frosted:
Maybe I just wanted to emphasise - POOOOOOSIX.

Anyways, thanks!

PR #744, opened 2024-10-30
with-poll flag is passed to configure

### LINKED ISSUE #738: [General]: with-poll not working
### Issue description

I can't seem to get `--with-poll` to work.  I've installed it with the flag, but it doesn't seem to carry over when compiling. It certainly could be user error. I'm not using a specific branch, maybe I should be? 

Also, I love emacs-plus, been using it for years. Thanks to all who contribute.

```
# uninstall first
brew uninstall emacs-plus@30
# install
brew install emacs-plus@30 --with-native-comp --with-poll
```

I've done some digging and it doesn't appear in the places I'd expect:

When I look at the config.log it doesn't appear there

```
It was created by GNU Emacs configure 30.0.91, which was
generated by GNU Autoconf 2.72.  Invocation command line was

  $ ./configure --disable-dependency-tracking --disable-silent-rules
 --enable-locallisppath=/opt/homebrew/share/emacs/site-lisp 
--infodir=/opt/homebrew/Cellar/emacs-plus@30/30.0.91/share/info/emacs 
--prefix=/opt/homebrew/Cellar/emacs-plus@30/30.0.91
 --with-xml2 --with-gnutls --with-native-compilation=aot 
--without-compress-install --without-dbus 
--without-imagemagick --with-modules --with-rsvg
 --with-webp --with-ns --disable-ns-self-contained
```

When I check via `C-h v system-configuration-options`

It shows:

```
"--disable-dependency-tracking --disable-silent-rules
 --enable-locallisppath=/opt/homebrew/share/emacs/site-lisp
 --infodir=/opt/homebrew/Cellar/emacs-plus@30/30.0.91/share/info/emacs
 --prefix=/opt/homebrew/Cellar/emacs-plus@30/30.0.91
 --with-xml2 --with-gnutls --with-native-compilation=aot 
--without-compress-install --without-dbus --without-imagemagick
 --with-modules --with-rsvg --with-webp --with-ns 
--disable-ns-self-contained 'CFLAGS=-O2 -DFD_SETSIZE=10000 
-DDARWIN_UNLIMITED_SELECT -I/opt/homebrew/opt/gcc/include
 -I/opt/homebrew/opt/libgccjit/include' 'LDFLAGS=-L/opt/homebrew/lib/gcc/14 
-I/opt/homebrew/opt/gcc/include -I/opt/homebrew/opt/libgccjit/include'"
```

And `system-configuration-features` doesn't show it. I think POLL should be in here:

"ACL GIF GLIB GMP GNUTLS JPEG LCMS2 LIBXML2 MODULES NATIVE_COMP NOTIFY KQUEUE NS PDUMPER PNG RSVG SQLITE3 THREADS TIFF TOOLKIT_SCROLL_BARS TREE_SITTER WEBP XIM ZLIB"

This isn't a huge deal unless I open a couple different projects using LSP.

aaronjensen:
Fixes #738

### DIFF
--- a/Formula/emacs-plus@29.rb
+++ b/Formula/emacs-plus@29.rb
@@ -157,6 +157,7 @@ class EmacsPlusAT29 < EmacsBase
     args << "--with-rsvg"
     args << "--without-pop" if build.with? "mailutils"
     args << "--with-xwidgets" if build.with? "xwidgets"
+    args << "--with-poll" if build.with? "poll"
 
     system "./autogen.sh"
 
@@ -250,7 +251,6 @@ class EmacsPlusAT29 < EmacsBase
         (man1/"ctags.1").unlink
       end
     end
-    args << "--with-poll" if build.with? "poll"
   end
 
   def post_install
--- a/Formula/emacs-plus@30.rb
+++ b/Formula/emacs-plus@30.rb
@@ -171,6 +171,7 @@ class EmacsPlusAT30 < EmacsBase
     args << "--with-webp"
     args << "--without-pop" if build.with? "mailutils"
     args << "--with-xwidgets" if build.with? "xwidgets"
+    args << "--with-poll" if build.with? "poll"
 
     system "./autogen.sh"
 
@@ -264,7 +265,6 @@ class EmacsPlusAT30 < EmacsBase
         (man1/"ctags.1").unlink
       end
     end
-    args << "--with-poll" if build.with? "poll"
   end
 
   def post_install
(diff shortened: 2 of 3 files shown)


aaronjensen:
Still working on testing this.

aaronjensen:
Tested with 31. From what I could see, this likely never worked.

d12frosted (approved):
good catch

d12frosted:
You don't need to pass it. We apply a patch:

```
  local_patch "poll", sha: "59e876f82e6fd8e4583bc2456339eda4f989c86b1e16a02b0726702e95f60825" if build.with? "poll"
```

or am I missing something?

d12frosted:
Oh. In that case we need to keep the option, but revert the deletion of the following line:

```
args << "--with-poll" if build.with? "poll"
```

aaronjensen:
@d12frosted not sure I understand... I moved the line because it was in the wrong place. Do we need it twice for some reason?

d12frosted:
Oh sorry, my bad. Looked at the wrong file 🤦 . Your fix seems legit - adding to args needs to be done BEFORE autogen / configure 😅 

And you are right, the commit that added this option never actually enable it 🤦

PR #746, opened 2024-11-01
Poll patch is removed

### LINKED ISSUE #745: [General]: with-poll causes issues with elpaca
### Issue description

I don't know the cause, and it is probably beyond my capacity and time to debug it, but now that with-poll works, I've tried using it and found that it appears to cause issues with asynchronous shell commands. Elpaca hangs on initial install. Magit appears to hang when issuing git commands.

If anyone else can corroborate this, I'd recommend removing the patch. It seems to do more harm than good.

aaronjensen:
It was not working from the day it was introduced and once it started working it started causing problems.

Fixes #745


d12frosted:
Agreed, let's remove it.

@aaronjensen thank you very much for dealing with this option 🙏

d12frosted (approved):

daviderestivo:
All emacs-head Formulas come with `FD_SETSIZE=10000` and `DARWIN_UNLIMITED_SELECT`. This is enough to make this patch obsolete.

```
./Formula/emacs-head@29.rb:    ENV.append "CFLAGS", "-O2 -march=native -pipe -DFD_SETSIZE=10000 -DDARWIN_UNLIMITED_SELECT"
./Formula/emacs-head@28.rb:    ENV.append "CFLAGS", "-O2 -march=native -pipe -DFD_SETSIZE=10000 -DDARWIN_UNLIMITED_SELECT"
./Formula/emacs-head@27.rb:    ENV.append "CFLAGS", "-O2 -march=native -pipe -DFD_SETSIZE=10000 -DDARWIN_UNLIMITED_SELECT"
./Formula/emacs-head@26.rb:    ENV.append "CFLAGS", "-O2 -march=native -pipe -DFD_SETSIZE=10000 -DDARWIN_UNLIMITED_SELECT"
./Formula/emacs-head@31.rb:    ENV.append "CFLAGS", "-O2 -march=native -pipe -DFD_SETSIZE=10000 -DDARWIN_UNLIMITED_SELECT"
```

PR #755, opened 2024-11-14
Update no title bar method link for Emacs 29+

maxattwell:
Readme update to fix the no title bar "this method" link in 29, 30 and 31 as I guess the title was updated with emacs 31.


d12frosted (approved):
thank you!

PR #760, opened 2024-11-26
Fix round-undecorated-frame patch for Emacs 31

mkvoya:
The round-undecorated-frame patch fails to apply for recent Emacs 31. This MR fixes the issue by regenerating the patch after a manual fix.

NOTE: this MR changes the "FROM" field of the patch. I don't know how to specify that; perhaps we should keep the original author field "Boris Buliga <boris@d12frosted.io>" unchanged.


mkvoya:
I just changed the patch author info back to @d12frosted .

d12frosted (approved):
Thanks a lot 🙏

PR #777, opened 2025-02-11
docs: delete the link `chunkwm` because it was returning a 404

struuuuggle:
For some reason, the repository named "[chunkwm](https://github.com/koekeishiya/chunkwm)" returned a 404 status, which is created by the same developers as [yabai](https://github.com/koekeishiya/yabai). I removed it from the README.

### DIFF
--- a/README.org
+++ b/README.org
@@ -232,7 +232,7 @@ And if for some reason PATH injection doesn't work for you, report it either in 
 
 **** Emacs 28 and Emacs 27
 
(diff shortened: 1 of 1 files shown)


d12frosted:
Hey. "some reason" is that Åsmund dropped support of chunkwm long time ago and eventually decided to remove the repository 😅 The doc was written when both were available.

Thanks

d12frosted (approved):

PR #784, opened 2025-02-24
Turn native comp on by default in emacs 30+

dustinfarris:
Emacs changes the default for this flag to 'yes'.  We've been changing that to 'aot' for previous versions.  This carries that opinion forward and removes the opt-in flag.

I have tested this locally.

### DIFF
--- a/.github/workflows/emacs-30.yml
+++ b/.github/workflows/emacs-30.yml
@@ -28,7 +28,6 @@ jobs:
         build_opts:
           - ""
           - "--with-xwidgets"
-          - "--with-native-comp"
           - "--with-x11"
 
     env:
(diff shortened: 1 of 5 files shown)


d12frosted (approved):
Well, one option less. Which is amazing in my book. Thanks a lot!

P.S. I do fear this change though as native comp is the source of such a huge instability, but hold tight and let's go!

dustinfarris:
> Well, one option less. Which is amazing in my book. Thanks a lot!
> 
> P.S. I do fear this change though as native comp is the source of such a huge instability, but hold tight and let's go!

I figure we can add an opt-out flag if the need presents itself.

jidicula:
@d12frosted I opened a PR to update the README to reflect this - is this needed? If not, feel free to reject and close it 🙂.

PR #802, opened 2025-03-05
Update post-install instructions to mv

### LINKED ISSUE #801: App alias doesn't show up in Spotlight
Is it really impossible to install Emacs.app into /Applications/Emacs.app directly? I install like this on macOS 15.3.1:

```sh
brew install emacs-plus@30 --with-modern-doom3-icon
osascript -e 'tell application "Finder" to make alias file to posix file "/opt/homebrew/opt/emacs-plus@30/Emacs.app" at posix file "/Applications" with properties {name:"Emacs.app"}'
```

and when I open Spotlight with Cmd+Space, and type "emacs", it doesn't show up. I have to scroll all the way down and it's under "Other", effectively adding like 3-4 seconds to Emacs's startup time. Does it really need to be installed into /opt/homebrew/opt/emacs-plus@30/Emacs.app and aliased? If you make a symlink with `ln -s` it doesn't show up at all.

verhovsky:
Is there a reason to alias the app instead of moving it? I tried this and emacs seems to still work fine and shows up in Spotlight this time.

Fixes #801

### DIFF
--- a/Formula/emacs-plus@31.rb
+++ b/Formula/emacs-plus@31.rb
@@ -267,8 +267,8 @@ class EmacsPlusAT31 < EmacsBase
       Emacs.app was installed to:
         #{prefix}
 
-      To link the application to default Homebrew App location:
-        osascript -e 'tell application "Finder" to make alias file to posix file "#{prefix}/Emacs.app" at posix file "/Applications" with properties {name:"Emacs.app"}'
+      To get Emacs to show up in Spotlight, move the app to the /Applications folder:
+        mv "#{prefix}/Emacs" /Applications/Emacs.app
 
       Your PATH value was injected into Emacs.app/Contents/Info.plist
 
--- a/Formula/emacs-plus@26.rb
+++ b/Formula/emacs-plus@26.rb
@@ -115,8 +115,8 @@ class EmacsPlusAT26 < EmacsBase
       Emacs.app was installed to:
         #{prefix}
 
-      To link the application to default Homebrew App location:
-        osascript -e 'tell application "Finder" to make alias file to posix file "#{prefix}/Emacs.app" at posix file "/Applications" with properties {name:"Emacs.app"}'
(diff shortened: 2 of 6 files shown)


verhovsky:
I see, `doom sync` doesn't work

Formula/emacs-plus@31.rb
```diff
@@ -267,8 +267,8 @@ def caveats
       Emacs.app was installed to:
         #{prefix}
 
-      To link the application to default Homebrew App location:
-        osascript -e 'tell application "Finder" to make alias file to posix file "#{prefix}/Emacs.app" at posix file "/Applications" with properties {name:"Emacs.app"}'
+      To get Emacs to show up in Spotlight, move the app to the /Applications folder:
+        mv "#{prefix}/Emacs" /Applications/Emacs.app
```
d12frosted, line 271: shouldn't it be 

```
mv "#{prefix}/Emacs.app" /Applications/Emacs.app
```

?

verhovsky: doesn't matter, it doesn't work anyway

verhovsky: it would be really nice if this could be fixed. I have had to open emacs by going to Finder, pressing Cmd+a to go to /Applications and typing "Ema" and pressing Cmd+o for the last few months and it's really annoying

d12frosted: Look, Emacs+ doesn't provide cask, meaning it can not install any `.app` files in `/Application` or `~/Application`. So for now it's up to the user to deal with that 🤷 

I have just tried, the following works:

```
cp -r /opt/homebrew/opt/emacs-plus@30/Emacs.app ~/Applications/Emacs.app
```

d12frosted: > doesn't matter

actually it does matter :) Because the following also works:

```
cp -r /opt/homebrew/opt/emacs-plus@30/Emacs.app /Applications/Emacs.app
```

Issue #804, opened 2025-03-23
[Build Failure]: Patching emacs-plus@31 fails


dertuxmalwieder:
### Version

emacs-plus@30

### Make sure to follow these steps before submitting the issue

- [x] run `brew update` and try to reproduce the issue again
- [x] run `brew doctor`, fix all issues and try to reproduce your issue again
- [x] try to reproduce your issue without any custom options to the formula

### What happened?

<pre>
$ brew install emacs-plus@31 --with-imagemagick --with-dbus --with-savchenkovaleriy-big-sur-3d-icon --with-xwidgets
</pre>

<details>
  <summary>Command output</summary>
  <pre>
==> Fetching d12frosted/emacs-plus/emacs-plus@31
==> Cloning https://github.com/emacs-mirror/emacs.git
Updating /Users/tux0r/Library/Caches/Homebrew/emacs-plus@31--git
==> Checking out branch master
Your branch is up to date with 'origin/master'.
HEAD is now at 92b373318d2 Reduce code duplication in ns_set_appearance
==> Installing emacs-plus@31 from d12frosted/emacs-plus
==> Patching
==> Applying fix-window-role.patch
==> Applying system-appearance.patch
Error: Failure while executing; `patch -g 0 -f -p1` exited with 1. Here's the output:
patching file 'src/frame.h'
patching file 'src/nsfns.m'
1 out of 1 hunks failed--saving rejects to 'src/nsfns.m.rej'
patching file 'src/nsterm.m'
1 out of 7 hunks failed--saving rejects to 'src/nsterm.m.rej'

If reporting this issue please do so at (not Homebrew/brew or Homebrew/homebrew-core):
  https://github.com/d12frosted/homebrew-emacs-plus/issues
  </pre>
</details>

### brew doctor output

Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry or file an issue; just ignore this. Thanks!

Warning: Some installed formulae are deprecated or disabled.
You should find replacements for the following formulae:
  openssl@1.1

Warning: You have the following deprecated, official taps tapped:
  Homebrew/homebrew-services
Untap them with `brew untap`.

Warning: The following casks have the same name as core casks:
  macos-fuse-t/cask/fuse-t
Their taps are in use, so you must use these full names throughout Homebrew.

Warning: The following formulae have the same name as core formulae:
  tgragnato/amule/terraform
[...]

PR #805, opened 2025-03-31
Emacs 31 build fixes

aaronjensen:
Fixes https://github.com/d12frosted/homebrew-emacs-plus/issues/804

I re-applied the system-appearance patch myself and made the new patches, rather than using the patches provided in the comments in that issue because the patch difference didn't look quite right. 

I tested the system appearance patch by running `emacs -Q` and toggling system appearance and confirming that it switched between a dark frame and a light frame. If that's not the way to test it, it should be tested.

### DIFF
--- a/Formula/emacs-plus@31.rb
+++ b/Formula/emacs-plus@31.rb
@@ -15,7 +15,6 @@ class EmacsPlusAT31 < EmacsBase
   option "without-cocoa", "Build a non-Cocoa version of Emacs"
 
   # Opt-in
-  option "with-ctags", "Don't remove the ctags executable that Emacs provides"
   option "with-x11", "Experimental: build with x11 support"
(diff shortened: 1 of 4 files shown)


d12frosted:
wow, @aaronjensen you are the hero ❤️

Formula/emacs-plus@31.rb
```diff
@@ -242,17 +241,6 @@ def install
       system "gmake"
       system "gmake", "install"
     end
-
-    # Follow MacPorts and don't install ctags from Emacs. This allows Vim
```
d12frosted, line 246: tbh, I don't remember why we added it
but a fair question - why did you remove it? :)

aaronjensen: Ctags support has been removed, so I don't think this is needed anymore.

d12frosted (approved):

PR #809, opened 2025-04-26
Add support for smooth cursor for emacs 31

mkvoya:
The smooth cursor feature adds native cursor movement animation for macOS. The feature extends [ksqsf's implementation](https://github.com/ksqsf/emacsmoe/pull/7) to support more (if not all) kinds of cursors and cursor blinks.

### DIFF
--- a/Formula/emacs-plus@31.rb
+++ b/Formula/emacs-plus@31.rb
@@ -20,6 +20,7 @@ class EmacsPlusAT31 < EmacsBase
   option "with-xwidgets", "Experimental: build with xwidgets support"
   option "with-no-frame-refocus", "Disables frame re-focus (ie. closing one frame does not refocus another one)"
   option "with-compress-install", "Build with compressed install optimization"
+  option "with-smooth-cursor", "Build with animated smooth cursor"
 
   #
   # Dependencies
@@ -93,6 +94,7 @@ class EmacsPlusAT31 < EmacsBase
   local_patch "fix-window-role", sha: "1f8423ea7e6e66c9ac6dd8e37b119972daa1264de00172a24a79a710efcb8130"
   local_patch "system-appearance", sha: "53283503db5ed2887e9d733baaaf80f2c810e668e782e988bda5855a0b1ebeb4"
   local_patch "round-undecorated-frame", sha: "26947b6724fc29fadd44889808c5cf0b4ce6278cf04f46086a21df50c8c4151d"
+  local_patch "ksqsf-smooth-cursor", sha: "c77f045e3defe30f7a8822b4d7962c37c84afcde9198d6131bf13894af2e72ea" if build.with? "smooth-cursor"
 
   #
   # Install
--- a/README.org
+++ b/README.org
@@ -142,6 +142,7 @@ By default =emacs-plus@31= uses the following features.
 | =--with-xwidgets=         | build [[#xwidgets-webkit][→ with xwidgets]] support                                                |
 | =--without-cocoa=         | build a non-Cocoa version of Emacs (terminal only)                           |
 | =--with-imagemagick=      | build with =imagemagick= support                                               |
+| =--with-smooth-cursor=    | build with smooth cursor support                                             |
 
 *** No title bar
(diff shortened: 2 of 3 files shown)


daviderestivo:
@mkvoya would you be interested in opening a PR for https://github.com/daviderestivo/homebrew-emacs-head as well :)?

d12frosted:
Hey @mkvoya 

Thanks for PR. Two questions:

1. Where did it come from? Are you the original author?
2. Did you propose it to upstream? If yes, can you share a link to communication? If not - why? :)

As a rule of thumb I try to minimise amount of included patches and I prefer people to contribute to the upstream, so wider audience can enjoy improvements. Besides, practically I lack time and knowledge to support non-trivial patches.

ksqsf:
Hi, I'm the original implementor of the idea, and the original patch is in my own Emacs fork (https://github.com/ksqsf/emacsmoe/pull/7). Recently @mkvoya improved it greatly, sent it to me, and we thought it might be a good idea to share it with greater audience.

1. @mkvoya can be considered a co-author at the very least, or a full author in its current form at best.
2. No, this feature will not be accepted by the upstream, because it cannot work on "free operating systems". So there's no communication, since we didn't even try. :)

(Also, to the best of my knowledge, it will be quite difficult to implement it for Linux. Maybe a fun challenge, but the gain does not justify the effort.)

> Besides, practically I lack time and knowledge to support non-trivial patches.

This is a very valid concern. My suggestion is to apply this patch to Emacs 30 (rather than HEAD), so that the maintenance burden is minimal.

mkvoya:
The burden of maintaining the patch should not be large, as it involves only code additions, and the patch is quite stable in my nearly two months of daily work. I can maintain the patch if necessary since I am using it myself. By the way, @ksqsf  would you please share your email address so that I can list you in the patch's co-authors?

ksqsf:
> The burden of maintaining the patch should not be large, as it involves only code additions, and the patch is quite stable in my nearly two months of daily work. I can maintain the patch if necessary since I am using it myself. By the way, @ksqsf would you please share your email address so that I can list you in the patch's co-authors?

gmail beginning with justksqsf

d12frosted:
Thank you for this contribution - I really appreciate the work you've put into it.

I'm genuinely torn on this. I love providing neat features to macOS users, and if upstream won't accept these improvements, it seems natural to offer them through Emacs+. However, my 10 years of experience maintaining this project has taught me what happens to complex patches that remain outside upstream: contributors come and go, patches break, and I simply don't have the capacity nor desire to maintain them indefinitely.

**On the maintenance burden:**

Whilst I don't doubt the patch's current quality or stability, two months of testing is unfortunately quite short compared to Emacs's development cycle. Even pinning to Emacs 31 only defers the problem - users will eventually request Emacs 32 support, and at some point the patch will likely break. I've seen this pattern repeatedly; even our minimal no-titlebar patch has broken upstream more than three times over the years.

**A potential path forwards:**

I've been thinking about Emacs+ extensibility more broadly (patches like this, plus custom icons), and I'm increasingly convinced we need a proper mechanism for applying local/remote patches. This would let me keep the core formula simple (Homebrew generally discourages excessive options, and I've already implemented several tricks to manage what we have) whilst still supporting community innovations.

My proposal: I'll develop a custom wrapper that makes it straightforward to apply additional patches, and then we can document your patch in the README as a community-maintained option. This way, interested users can easily adopt it, but without committing the project to long-term maintenance of code that may not survive upstream changes.

What do you think of this approach?

d12frosted:
ok, this is what I mean - https://github.com/d12frosted/homebrew-emacs-plus/issues/851

I have something like a design document that I can share if someone is interested, but the rough idea is pretty simple;

implementation will come only after I deal with #783 as both `Emacs Client.app` and #851 both require to change a few formula internals and I don't want to deal with conflicts

mkvoya:
The mechanism for additional patches is great. Thanks for your invaluable effort!

d12frosted:
described mechanism implementation is mostly ready - see https://github.com/d12frosted/homebrew-emacs-plus/pull/858 for more information. I still need to test a few edge cases before I can merge it, but if you want to give it a shot with this patch - feedback is more than welcome 🙏

d12frosted:
closing this PR as I am not going to merge it in current state. If you want, convert it intro community patch and distribute it as described in relevant guides.

Issue #818, opened 2025-07-31
[General]: Build fails for emacs31


pjj:
### Issue description

As of today, Emacs top of tree applied the patch fix-window-role, and therefore the recipe to apply this same patch fails:

% brew reinstall emacs-plus@31 --with-dbus
==> Fetching downloads for: emacs-plus@31
==> Fetching d12frosted/emacs-plus/emacs-plus@31
==> Cloning https://github.com/emacs-mirror/emacs.git
Updating /Users/XXX/Library/Caches/Homebrew/emacs-plus@31--git
==> Checking out branch master
Already on 'master'
Your branch is up to date with 'origin/master'.
HEAD is now at 6e1054a40bf NS: Correct the accessibility role of Emacs windows (bug#77062)
==> Reinstalling d12frosted/emacs-plus/emacs-plus@31 --with-dbus --with-modern-orange-icon
==> Patching
==> Applying fix-window-role.patch
Error: Failure while executing; `patch -g 0 -f -p1` exited with 1. Here's the output:
patching file 'src/nsterm.m'
1 out of 1 hunks failed--saving rejects to 'src/nsterm.m.rej'

hazardfn:
Anyone who desperately needs a build can use something like `HOMEBREW_EMACS_PLUS_31_REVISION=a6e0bb92bb0c114fc8d57864455330d04eac064e brew install emacs-plus@31 --with-dbus --with-mailutils --with-xwidgets --with-imagemagick`

This builds Emacs 31 using the commit before the one that breaks the install: https://github.com/emacs-mirror/emacs/commit/6e1054a40bf6df1429a2b16fdd0d7652dae4d537

Latest commit that builds: https://github.com/emacs-mirror/emacs/commit/a6e0bb92bb0c114fc8d57864455330d04eac064e

Bit redundant now, but if something like this happens again, you find yourself here and you don't want to wait for a merge/fix you can always use `brew edit` to alter the formula temporarily.

d12frosted:
Merged fix by @aaronjensen

PR #819, opened 2025-08-01
Renable emacs-plus@31 installation

pjj:
This fixes: https://github.com/d12frosted/homebrew-emacs-plus/issues/818


dustinfarris:
@d12frosted can we get this in?

d12frosted:
Thanks for raising and PR. Fix is in master.

PR #820, opened 2025-08-03
Emacs 31 fix-window-role patch is removed

aaronjensen:
This was applied upstream

Fixes: #818 
Dupe of #819 (but has a single commit, rather than multiple)


d12frosted:
Thank you for a PR that was easy to merge from phone 🙃 

PS. Apologies, I am not very responsive. Life 🤷

d12frosted:
Oh and thanks for TLDR for breakage reason

aaronjensen:
You bet, thank you for the merge.

PR #825, opened 2025-08-15
Emacs 30.2

dotemacs:
Updated the formula for Emacs 30.2 install, since it was released


dotemacs:
> thanks a lot; let's wait for CI before merging

That's exactly what I'm doing, watching the CI builds 😄

PR #833, opened 2025-09-27
dev: allow installing from paths on CI and in dev tools

d12frosted:


d12frosted:
ooof, I need to refactor tests and check icons in a different way

d12frosted:
seems to pass, at least I fixed what I wanted to fix initially
will tackle logs in a different PR

PR #855, opened 2025-12-10
Update emacs-plus@30 dependencies

brorbw:
`ts_language_version` is deprecated in version 0.26 so use 0.25.

I'm currently without a working `emacs` - I'll fix the commits when it works.

### DIFF
--- a/Formula/emacs-plus@30.rb
+++ b/Formula/emacs-plus@30.rb
@@ -51,7 +51,7 @@ class EmacsPlusAT30 < EmacsBase
   depends_on "gnutls"
   depends_on "librsvg"
   depends_on "little-cms2"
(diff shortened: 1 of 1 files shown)


d12frosted:
ok, that's a good workaround actually, but let's see if my patch with shim works - it would be a bit safer than pinning to old version of tree-sitter

d12frosted:
seems like #856 works nicely
thanks a lot for PR let's stick to the patch I've created based on the upstream fix

brorbw:
This was the most minimal change I could muster without a working editor so thank you for making a more robust solution! It's very appreciated!

d12frosted:
ok seems it's not as robust; to avoid breakage, I am merging your PR :)

PR #866, opened 2025-12-24
Adds window blurring and alpha transparency patch to community patches

aaratha:
## This PR adds support for window blurring and alpha transparency.

+ Alpha transparency logic is largely inspired by @jaaasonSun’s patch in his fork of this repository.

+ Window blurring is implemented using private CGS APIs, following approaches used in Ghostty, Terminal.app, and other macOS terminals.

## Notes:

Both features are opt-in via frame parameters (alpha and ns-background-blur).

Blur only works on Cocoa frames and is applied dynamically. You may notice that it doesn’t appear until the frame is updated, for example, when opening the command buffer.

### DIFF
--- a/community/patches/aggressive-read-buffering/README.md
+++ b/community/patches/aggressive-read-buffering/README.md
@@ -1,0 +1,32 @@
+# aggressive-read-buffering
+
+Improve processing large quantities of data from tty/pty reads on macOS
+
+## Compatibility
+
+- Emacs versions: 29, 30, 31
+
+## Maintainer
+
+aport
+
+## Usage
+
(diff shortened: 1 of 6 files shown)


aaratha:
Yeah, I guess I should've realized the macOS part was redundant lol. I'll fix that and make it explicit that it's a frame modification, not window. Thanks for the note.

d12frosted:
there is a merge conflict 😭 can you resolve it, so I can merge? 🙏

d12frosted:
I will be near my personal laptop in a few hours though, if you don't resolve it by then - I will do myself.

aaratha:
sorry I didn't see this. Must have been asleep lol

PR #867, opened 2025-12-25
community: add patch to improve async process read performance

aport:
On macOS the maximum buffer size of a tty/pty is 1KB. When reading the output data from an async process, the function emacs_intr_read() only issues a single call to read() and then passes the data up the stack for processing. This effectively limits the value of read-process-output-max to 1024 bytes on macOS.

If we've read 1KB on a TTY, it's highly likely we've read the maximum and the writer has more data to send. Continue reading in a loop, with a short delay, to allow the writer to put more data. This lets us bulk read data (up to read-process-output-max bytes) in a single call to emacs_read().

On a MBP with an M4 Pro, running in the emacs git repository:

`(async-shell-command "time git -C ~/emacs --no-pager log --oneline")`

Before:
1.23s user 0.19s system 1% cpu 2:21.32 total

After:
0.76s user 0.07s system 53% cpu 1.550 total

Same command in iTerm2 for comparison:
0.90s user 0.15s system 98% cpu 1.076 total

### DIFF
--- a/community/patches/aggressive-read-buffering/README.md
+++ b/community/patches/aggressive-read-buffering/README.md
@@ -1,0 +1,32 @@
+# aggressive-read-buffering
+
+Improve processing large quantities of data from tty/pty reads on macOS
+
+## Compatibility
+
+- Emacs versions: 29, 30, 31
+
+## Maintainer
+
+aport
+
+## Usage
+
+Add to your `~/.config/emacs-plus/build.yml`:
+
+```yaml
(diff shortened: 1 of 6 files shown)


d12frosted:
wow, that's nice; I wonder if upstream would accept this patch;
meanwhile, it does sound good enough to be included as community patch here, so many thanks ❤️

d12frosted (approved):

PR #868, opened 2025-12-26
adds configurable element transparency, updates to latest emacs version

aaratha:
Fixes nsterm.m hunk rejection

### DIFF
--- a/community/patches/frame-transparency/README.md
+++ b/community/patches/frame-transparency/README.md
@@ -30,7 +30,27 @@ Then rebuild Emacs:
 brew reinstall emacs-plus@31
 ```
 
-To use the new features, add alpha transparency and blur radius parameters in your config:
+To use the new features, choose which elements should be rendered with transparency
+
+``` emacs-lisp
+;; None:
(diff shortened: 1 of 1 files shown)


jwintz:
This patch is awesome. However, it blends all background faces, which includes highlights (current region, `hl-line-mode` etc.). So for alpha = 0, they're not visible. I would only apply the alpha factor to the background attribute for the 'default face, and handle window background and fringes.

aaratha:
Yeah @jwintz, I want to cut down on the number of changes too so hopefully the patch works longer. I'll work on that.

aaratha:
I ended up making the patch a bit more complex, with opt-in options available for major elements. While this improves customizability, it will likely make maintenance more difficult in the long run. I’m curious to hear what people think.

Also, regarding the comment above about hl-line, it seems that those backgrounds are determined by the more general glyph background fill, so separating that logic would require more specific checks. Otherwise, disabling transparency for glyphs should reverse this behavior.

jwintz:
Excellent. Works perfectly. Thanks for the contribution!

PR #880, opened 2025-12-30
Fix tree-sitter compatibility for Emacs 29 and 30

daviderestivo:
Replace ts_language_version with ts_language_abi_version to support newer tree-sitter versions that removed the deprecated function.

- Add treesit-compatibility.patch for Emacs 29 and 30
- Apply patches conditionally when tree-sitter option is enabled
- Use local_patch method for proper integration

Based on commit c27c6f37126fd6045983ce91aa4118b42849d6f2 from daviderestivo/homebrew-emacs-head


d12frosted:
Oh thank you. This also means that I can remove tree sitter version pin.

PR #894, opened 2026-01-07
docs(readme): Add build from source and Centaur Emacs links

seagle0128:
Added new instructions for building from source (Formula) and included a link to Centaur Emacs as an inspiring project.


d12frosted:
Thanks ❤️

PR #938, opened 2026-04-23
Updates patch to latest build

aaratha:
Edits patch to be compatible with latest builds of emacs and emacs-plus.


d12frosted:
Hi @aaratha — thanks for spotting and fixing this! Confirmed the context mismatch: upstream Emacs master now has an `ApplicationServices` comment block + `#import` between `#include <IOSurface/IOSurface.h>` and `#endif` in `src/nsterm.m`, and the old hunk context no longer matches. Your expanded context is exactly right.

I've cherry-picked your patch-file change directly to master in c4c0c2b with authorship preserved — so your contribution is in, thank you!

Apologies for the noise this PR accumulated: the `.DS_Store` and the `Casks/emacs-plus-app.rb` downgrade (plus the `github-actions[bot]` follow-up commit) were caused by our `build-app.yml` workflow running on your fork when you pushed to master, producing a fork-local release and auto-committing a cask rewrite pointing at it. That was our bug, not yours. I've guarded the workflow with a `github.repository` check in fd3a196 and added `.DS_Store` to `.gitignore` in 1c00f6c so this doesn't happen to the next contributor.

Going to close this PR since the fix is already merged. Thanks again!

aaratha:
Great! Glad I could help, and I appreciate the authorship note.

PR #940, opened 2026-05-02
updates readme to include example config

aaratha:
I managed to edit my config to fix a common issue with the blur effect not appearing on startup. I changed my readme format to org-mode and included an example config that seems to work in the cases I've tested.


d12frosted (changes requested):
Thanks for this! The example config is genuinely useful - the blur-on-startup workaround is a real pain point and worth documenting. A few things I'd like to sort out before merging:

**1. Please keep the README in Markdown.**

The [community contribution guide](../blob/master/community/README.md) requires `README.md`, and the other community patches (`aggressive-read-buffering`, `mac-font-use-typo-metrics`) all use Markdown too. I'd rather not have one patch diverge in format. Could you append the new "Example Config" section to the existing `README.md` instead of converting the file?

**2. Duplicated block at the end of the example.**

```elisp
(when (display-graphic-p)
  (my/apply-frame-transparency))
(when (display-graphic-p)
  (my/apply-frame-transparency))
```

Looks like a copy/paste slip - the same `when` block is repeated. Should just be one.

Once those are addressed I'm happy to merge. Thanks again for taking the time to write this up!

aaratha:
Thanks for the comments. I forgot about the markdown requirement. I've added the changes you suggested. Sorry for the extraneous commits.

d12frosted:
thanks a lot! merged

PR #941, opened 2026-05-09
Emacs 32 build target

aaronjensen:


aaronjensen:
@d12frosted I think this is ready to go, but please check to see if I missed anything.

d12frosted:
thanks a lot ❤️

PR #947, opened 2026-06-10
Disable MacOS 26 Tahoe "AutoFill (Emacs)" process

avit:
On MacOS Tahoe, I observed a running process called "AutoFill (Emacs)" alongside the Emacs app. That seems unnecessary for regular Emacs usage since it's intended for one-time codes.

This adds a plist option to make autofill default to off. Since adding it I haven't seen that process reappear.

See: https://developer.apple.com/documentation/bundleresources/information-property-list/nsautofillrequirestextcontenttypeforonetimecodeonmac

(AI-assisted solution)

### DIFF
--- a/Library/EmacsBase.rb
+++ b/Library/EmacsBase.rb
@@ -640,6 +640,10 @@ class EmacsBase < Formula
     system "/usr/libexec/PlistBuddy -c 'Set NSMicrophoneUsageDescription Emacs requires permission to access the Microphone.' '#{plist}'"
     system "/usr/libexec/PlistBuddy -c 'Add NSSpeechRecognitionUsageDescription string' '#{plist}' || true"
     system "/usr/libexec/PlistBuddy -c 'Set NSSpeechRecognitionUsageDescription Emacs requires permission to handle any speech recognition.' '#{plist}' || true"
+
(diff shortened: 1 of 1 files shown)


d12frosted:
Thanks, this looks good! I checked the key against Apple docs and also confirmed that upstream Emacs does not set it in nextstep/templates/Info.plist.in (neither master nor emacs-30), so this is a genuine gap and not redundant. Since inject_protected_resources_usage_desc is called by all five formulas, every version gets it.

The CI failure is unrelated - master fails the same way right now (cask loading / gdk_pixbuf_query_loaders infra issue), so not blocking on it.

Merging as is. Two small follow-ups we will handle on our side:

1. Use the plist_set helper (add-or-set) instead of a bare Add, so the build does not break if upstream Emacs ever ships this key in its template.
2. Rename the method, since it is no longer only about usage descriptions.

Also planning to prepare a patch for upstream Emacs (nextstep/templates/Info.plist.in) so released versions eventually get this out of the box.

d12frosted:
FYI, added this to upstream - https://github.com/emacs-mirror/emacs/commit/8486669e373fb8e9962428c42b427462cc655e3c

PR #949, opened 2026-06-10
fix(ci): update Homebrew before installing formulas

d12frosted:
Build and Nightly workflows are currently red on every run with:

    Error: unknown install step: gdk_pixbuf_query_loaders

Root cause is version skew between the runner image and homebrew-core:

- Homebrew added a declarative install steps framework on 2026-05-22 and the gdk_pixbuf_query_loaders step handler on 2026-06-02 (shipped in brew 5.1.15, released 2026-06-03).
- gdk-pixbuf 2.44.6_1 in homebrew-core now declares its post-install through that framework, and the JSON API serves this metadata to all clients regardless of brew version.
- The macos-latest runner image bundles an older brew (the traceback points at install_steps.rb:244, which matches the 2026-05-22 framework version without the gdk_pixbuf handler), and HOMEBREW_NO_AUTO_UPDATE is set on runners, so brew never self-updates.

Result: installing any formula that depends on gdk-pixbuf (i.e. all emacs-plus formulas via librsvg) fails in the dependency post-install phase.

Fix: run brew update before installing, same as build-app.yml already does (which is why cask builds are not affected by this; their current failure is an unrelated transient Savannah 504 on git clone).

Also passes --formula to brew install when installing from a path, which silences the bogus 'Failed to load cask: Formula/emacs-plus@30.rb' error annotation in logs.

This PR touches build.yml, so the Build check on this PR itself verifies the fix.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,10 +29,18 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v6
 
+      # The runner image ships a stale Homebrew, while formula metadata
+      # comes fresh from the JSON API. When homebrew-core starts using a
+      # feature the bundled brew does not know yet (e.g. the install
+      # steps framework), dependency installation fails with errors like
+      # "unknown install step: gdk_pixbuf_query_loaders".
+      - name: Update Homebrew
+        run: brew update
+
       - name: Build emacs-plus@30
         run: |
           export HOMEBREW_DEVELOPER=1
-          brew install Formula/emacs-plus@30.rb --verbose
+          brew install --formula Formula/emacs-plus@30.rb --verbose
 
       - name: Test installation
         run: $(brew --prefix)/bin/emacs --batch --eval='(print (+ 2 2))'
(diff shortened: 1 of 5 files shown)


d12frosted:
The brew update fix worked - all 23 dependencies including gdk-pixbuf now install fine - but the newer brew surfaced a second breakage right after, in the Patching phase:

    Errno::ENOENT: No such file or directory @ rb_sysopen - /private/tmp/homebrew-sandbox.../patches/emacs-30/fix-window-role.patch

Since brew 5.1.15 formulas are loaded inside the build sandbox, whose working directory is a temp dir, so the Dir.pwd fallback in UrlResolver / EmacsBase.init resolved local patch paths against the sandbox instead of the checkout.

Pushed a second commit that derives the repo root from the location of UrlResolver.rb itself (Library/..), which works for checkout installs, make formula-XX local mode and tap installs alike. Added tests/test_url_resolver.rb that reproduces the sandbox cwd scenario (fails on the old code, passes now).

PR #959, opened 2026-06-21
Doc: Add Brew trust instruction

Nuno69:
Since version 6.0, Homebrew requires explicit trusting of newly added taps, formulae or casks.
This change adds the command the user needs to type in the terminal to trust the whole tap, however I am pondering whether its better to instruct users to trust only the specific formulae?

### DIFF
--- a/README.org
+++ b/README.org
@@ -24,6 +24,7 @@ Emacs+ is [[https://www.gnu.org/software/emacs/emacs.html][→ GNU Emacs]] formu
 
 #+begin_src bash
   $ brew tap d12frosted/emacs-plus
+  $ brew trust d12frosted/emacs-plus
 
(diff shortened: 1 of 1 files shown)


d12frosted:
thanks for the heads up on this! tap trust caught me off guard too.

i pushed a small follow-up commit on top of yours that reframes it a bit. instead of an unconditional `brew trust` line up front, it's now a note that only kicks in if you actually hit an untrusted tap error - mostly because the literal command errors out on older Homebrew (< 6.0), and i'd rather not hand folks a copy-paste trust step that defeats the whole point of the deliberate-trust prompt. it also mentions formula/cask-level trust and links the Tap-Trust docs for anyone who wants to read more.

merging now with rebase so your commit stays as-is. appreciate it!

Issue #733, opened 2024-09-23
[General]: libgccjit.so error invoking gcc driver


elondres-mim:
### Issue description

After updating to Sequoia I get the titular error every time I launch Emacs Plus from the applications folder or a saved-to-dock Emacs Plus and then run M-x.

I do *not* get the error in question when launching Emacs Plus from the command line, interestingly enough.

d12frosted:
Pretty sure it's related to https://github.com/d12frosted/homebrew-emacs-plus/issues/720

Env injection is broken for whatever reason. Frankly, I will not have any time soon to investigate, so hopefuly as more people upgrade, someone will be able to spend some time investigating. 

I wonder if using [exce-path-from-shell](https://github.com/purcell/exec-path-from-shell) solves the issue.

efroemling:
> Pretty sure it's related to #720

This is the exact problem I ran into which led me to looking into the paths thing, so I'd assume the same.

Prior to relying on emacs-plus' automatic path stuff I was using exec-path-from-shell so I'd imagine that could be a good fallback here too.

Though in my case I'm just explicitly setting the value from the Info.plist via a few lines at the top of my .emacs so that I theoretically get exactly what emacs-plus would be giving me:
``` lisp
;; Temp: Explicitly set PATH environment variable and update exec-path to match it.
;; (the string here should be copied from the PATH in Emacs.app/Contents/Info.plist)
(setenv "PATH" "/opt/homebrew/bin:/opt/homebrew/sbin:...etc etc...")
(setq exec-path (split-string (getenv "PATH") path-separator))
```

citizensn1ps:
> ```commonlisp
> ;; Temp: Explicitly set PATH environment variable and update exec-path to match it.
> ;; (the string here should be copied from the PATH in Emacs.app/Contents/Info.plist)
> (setenv "PATH" "/opt/homebrew/bin:/opt/homebrew/sbin:...etc etc...")
> (setq exec-path (split-string (getenv "PATH") path-separator))
> ```

@efroemling Thanks, this solution worked for me as well. Which kinda surprised me because the native compilation log was saying the libgccjit library emutls_w wasn't found. And only when compiling a handful of my packages. Sent me into the depths of the gcc directory...this solution is...much simpler 😅

al3xandru:
In my environment, setting the PATH as per efroemling's instructions did not work.
I got the impression that it worked, but then noticed again the exact same error.

[...]

Issue #562, opened 2023-04-09
Solution to native-comp installation on Apple Silicone


catesandrew:
I've failed and succeeded installing emacs-plus with native comp over the past half year. This spans various versions of emacs-plus, gcc, and liggccjit. Sometimes it works, sometimes it doesn't. I've posted answers to some of these issues that worked in the past. I've gone through the same issues it seems numerous times, looking for a fix.

Once again, today I was unable to install emacs-plus@30. Today, I couldn't even install gcc from source, let alone libgccjit. I tried in vein looking for a solution and came across [this gist](https://gist.github.com/cpbotha/5f44a7f44ad93445128113c208b4a252). This lent to a solution that worked for me to successfully install emacs-plus@30 with native comp.

I ended up grabbing the latest formulae from homebrew. Then commented out the stable section that patches the libraries. THen I updated the head to point to the actual repo and branch of the gcc darwin maintainer (vs patching). Finally, I removed the with-bugurl parameter. This allowed me to install gcc and libgccjit from source with no issues or errors about undefined method `issues_url' for nil:NilClass. Everything worked fairly easy afterwards. 

I'll remove this issue in a couple days once its fixed upstream, but for those stuck today, I hope it helps you too or inspires you for another solution.

```
brew tap catesandrew/tap
brew uninstall --ignore-dependencies gcc libgccjit
brew install --HEAD catesandrew/tap/gcc
brew install --HEAD catesandrew/tap/libgccjit

# CPATH and LIBRARY_PATH are necessary for libgccjit library discovery
HOMEBREW_NO_INSTALL_CLEANUP=1 \
  CPATH="${HOMEBREW_PREFIX}/include" \
  LIBRARY_PATH="${HOMEBREW_PREFIX}/lib/gcc/current" \
  brew install \
  --ignore-dependencies \
  emacs-plus@30 \
  --with-native-comp \
  --with-xwidgets \
  --with-imagemagick \
  --with-mailutils \
  --with-poll \
  --with-no-frame-refocus \
```

My `HOMEBREW_PREFIX` is `/opt/homebrew`, yours may be that or `/usr/local`.

Issue #554, opened 2023-02-26
Native comp doesn't work after update to MacOS 13.2.1 , `ld: library not found for -lemutls_w`


slinkp:
After a system update from 13.1.something to 13.2.1, emacsplus 28.1 with native comp enabled was failing to load [my config](https://github.com/slinkp/dotemacs).

(Preamble on the yak shaving I had to do to get this far:
I didn't save the errors but they appeared to be failing to load native compiled packages.
`brew doctor` told me I needed to reinstall xcode command line tools, so I did that.
Then I did `brew update && brew upgrade`.
Still could not load packages; I noticed `straight` was failing to fetch packages with git failures in `*Warnings*`. I realized my git auth was broken after the system update. Fixed that via `gh auth setup-git`.)

Still got many failures on startup, apparently during `straight` bootstrapping, like this:
```
ld: library not found for -lemutls_w
libgccjit.so: error: error invoking gcc driver
```

I noticed that emacs 28.2 was available, so I figured I'd try an upgrade and rebuild via:
```
brew uninstall emacs-plus@28
brew install emacs-plus@28 --with-native-comp --with-modern-papirus-icon
```

This apparently succeeded building, but when starting emacs, I saw the same errors.
```
ld: library not found for -lemutls_w
libgccjit.so: error: error invoking gcc driver
```

After some googling led to [this apparently identical error with doomemacs](https://github.com/doomemacs/doomemacs/issues/6789),
I founda [this comment with a workaround](https://github.com/doomemacs/doomemacs/issues/6789#issuecomment-1249606286) - if I start emacs inside the console:
```
emacs -nw
```
... This appears to complete native compilation successfully, and now I can start with the GUI as well.

Here is output of `brew doctor -v`:

```
==> Cask Environment Variables:
BUNDLE_PATH
CHRUBY_VERSION
GEM_HOME
GEM_PATH
HOMEBREW_CASK_OPTS
LC_ALL
PATH
RBENV_VERSION
RUBYLIB
RUBYOPT
RUBYPATH
SHELL
==> $LOAD_PATHS
/usr/local/Homebrew/Library/Homebrew/vendor/bundle/ruby/2.6.0/gems/warning-1.3.0/lib
/usr/local/Homebrew/Library/Homebrew/vendor/bundle/ruby/2.6.0/gems/tapioca-0.7.3/lib
[...]

PR #963, opened 2026-06-25
ci(build-app): fix libgccjit native-comp build on gcc 16 / macOS arm64

d12frosted:
## What

Make `./configure --with-native-compilation` find the gcc runtime that gcc 16's libgccjit driver links against, across all three build jobs:

- Install `gcc` explicitly and derive the versioned binary from `brew list --versions gcc` (no fragile globbing).
- Locate `libemutls_w.a` via `gcc -print-file-name` (the driver's own mechanism — deterministic, unlike a Cellar-wide `find | head -1`) and put that dir, the gcc lib dir, and the libgccjit lib dir on `LIBRARY_PATH`.
- Pin `SDKROOT` to the installed SDK, mirroring Homebrew superenv.
- On failure, surface `config.log` (grep + artifact) to make libgccjit linker errors debuggable in CI.

## Why

After the gcc 16.1.0 bottle landed (~2026-06-19), every arm64 `Build Emacs.app` job failed at the libgccjit configure test:

```
configure: error: The installed libgccjit failed to compile and run a test program
```

config.log showed the real cause — gcc 16's libgccjit driver links the JIT test with `-lemutls_w` and the linker couldn't find it:

```
ld: library 'emutls_w' not found
libgccjit.so: error: error invoking gcc driver
```

This is the recurring native-comp-on-macOS issue (#554, #733, #562). The earlier `LIBRARY_PATH` attempt (#962) used a flaky `find` that resolved inconsistently across runners (sequoia flipped pass/fail between identical runs); deriving the path from gcc itself makes it reliable.

## Validation

Manual `workflow_dispatch` (version 31) — all four platforms green, including tahoe/macOS 26 which never built before:

| Platform | Result |
|---|---|
| macos-26 arm64 (tahoe) | ✅ |
| macos-15 arm64 (sequoia) | ✅ |
| macos-14 arm64 (sonoma) | ✅ |
| macos-15-intel x86_64 | ✅ |

Follow-up to #961 and #962.

### DIFF
--- a/.github/workflows/build-app.yml
+++ b/.github/workflows/build-app.yml
@@ -71,7 +71,7 @@ jobs:
         run: |
           brew update
           brew install autoconf automake texinfo gnutls librsvg libxml2 \
-            libgccjit tree-sitter@0.25 pkg-config webp
+            gcc libgccjit tree-sitter@0.25 pkg-config webp
           # tree-sitter@0.25 is keg-only, set up paths
           echo "PKG_CONFIG_PATH=$(brew --prefix tree-sitter@0.25)/lib/pkgconfig:$PKG_CONFIG_PATH" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


lrusso96:
@d12frosted thanks for the fix. Do you also plan to update the stable cask with this change?

d12frosted:
yes, the fix lives in the shared build workflow, so the next stable (emacs 30) build will include it automatically. the current stable build (215) was made before gcc 16 landed, so the app itself is fine.

note the CI fix only affects building the app though. the part that matters on user machines is native compilation at runtime, and that is handled at install time by the tap, not by the app build. #965 and #966 improve exactly that, and both are merged now, so `brew update && brew reinstall --cask emacs-plus-app` picks it up, no new build needed.

PR #1011, opened 2026-09-19
Remove Emacs 32 scroll patch

aaronjensen:
From what I can tell, this is now in master but not in Emacs-31 yet, so I left it there


d12frosted (approved):
cool, thanks a lot
CI is failing for an unrelated reason

